Issue #2, opened 2015-12-09
Add after_connect feature to db_connection


josevalim:
We should do this here or in Ecto. Keep in mind we don't need to keep Ecto API, we could for example ask a tuple of `{Mod, fun, args}` to be given on configuration.

fishcakez:
I think we could use `Task.Supervisor.start_child` to start a task that reuses `DBConnection.run/3` by creating a fake `DBConnection.Task` pool module. That should allow Ecto to use its query API on on_connect, instead of the drivers.

I think its ok to accept `{mod, fun, args} | ((DBConnection.t) -> any)` like a Task. If you want something different let me know.

josevalim:
@fishcakez honestly, I don't care about folks using ecto API in after_connect. If you need to use it, it will be mostly a db specific command anyway, so you may as well go through the driver. So feel free to do the simplest thing :)

PR #4, opened 2015-12-13
Introduce Ecto-style API

fishcakez:
Firstly we provide a mechanism to re-use prepared queries on multiple connections in the same pool. This means that Ecto (or another library) can use its owns global cache of prepared queries with a pool and if a query has not been prepared on the checked out connection it can be prepared and executed.

A naive driver can use this mechanism by trying to execute the query and return `{:prepare, state}` when the database says the query is not prepared. A driver could handle its own cache per connection if it proved beneficial.

Secondly we move query parameters outside of the query term so that it is easier to combine prepared queries and executed parameters.

This means that Ecto (or another library) can do the equivalent of:

```elixir
{pool, opts} = repo.__pool__
case repo.fetch_query(key) do
  {:ok, query} ->
    DBConnection.execute(pool, query, params, opts)
  :error ->
    query = DBConnection.prepare!(pool, query, opts)
    repo.put_query(key, query)
    DBConnection.execute!(pool, query, params, opts)
end
```

Ecto also requires #2 and perhaps other changes.

### DIFF
--- a/examples/db_agent/lib/db_agent.ex
+++ b/examples/db_agent/lib/db_agent.ex
@@ -12,22 +12,19 @@ defmodule DBAgent do
   @spec get(DBConnection.conn, ((state :: any) -> value), timeout) ::
     value when value: var
   def get(conn, fun, timeout \\ 5_000) when is_function(fun, 1) do
-    DBConnection.query!(conn, {:get, fun},
-      [queue_timeout: timeout, prepare: :manual, decode: :manual])
+    DBConnection.query!(conn, :get, [fun], [queue_timeout: timeout])
   end
 
   @spec update(DBConnection.conn, ((state :: any) -> new_state :: any), timeout) ::
     :ok
   def update(conn, fun, timeout \\ 5_000) when is_function(fun, 1) do
-    DBConnection.query!(conn, {:update, fun},
-      [queue_timeout: timeout, prepare: :manual, decode: :manual])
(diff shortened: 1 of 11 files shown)


josevalim:
@fishcakez so this still requires the user to explicitly call prepare?

The approach I had in mind would simply pass a "reference" to the adapter and the adapter would take care of it. My concern with doing it explicitly is that now we need to call the pool/connection twice, no?

fishcakez:
@josevalim we can call the pool/connection with:
```elixir
{:ok, result} = DBConnection.run(pool, fn(conn) ->
   query = DBConnection.prepare!(conn, query, opts)
    repo.put_query(key, query)
    DBConnection.execute!(conn, query, params, opts)
end)
result
```
This is one checkout from the pool. `run/3` is like `transaction/3` except no begin/commit/rollback. The connection process is never called in DBConnection, state remains in the client process throughout `run/3`.

I wanted to try this approach first because the code is simple. Can you give an example of how you'd like to prepare?

josevalim:
@fishcakez sounds good!

fishcakez:
We could provide:
```elixir
@spec prepare_execute(conn, query, params, opts) :: {:ok, query, result} | {:error, err}
```
This would delay decoding the result of execute until after check in, whereas decoding blocks the connection with the `run/3` example (though there is no reason an adapter couldn't turn off decoding and do it manually).

PR #5, opened 2015-12-14
Introduce after_connect option

### LINKED ISSUE #2: Add after_connect feature to db_connection

fishcakez:
Closes #2


josevalim:
:shipit:

PR #6, opened 2015-12-15
Introduce DBConnection.Sandbox

fishcakez:
`DBConnection.Sandbox` is equivalent to `Ecto.Adapters.SQL.Sandbox` though requires a different adapter API. A nice advantage of this implementation is that sandboxing does not change the pooling or transaction code and is a wrapper around a `DBConnection` module. It might be possible to combine it with an ownership based pool.

### DIFF
--- a/integration_test/cases/backoff_test.exs
+++ b/integration_test/cases/backoff_test.exs
@@ -24,8 +24,10 @@ defmodule BackoffTest do
     assert_receive {:hi, ^conn}
 
     assert [
-      connect: [[_, _ | ^opts]],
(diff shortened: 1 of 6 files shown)


josevalim:
It looks great. A couple questions:

1. Does it mean the adapters themselves now need to implement the sandbox behaviour?

2. Is Sandbox the proper name? I think sandbox is one possible use case for this, it feels like a Proxy or something of sorts.

The name Proxy will be specially relevant if we can implement it orthogonally to the given module (i.e. if 1 is false).

PR #7, opened 2015-12-15
Introduce :proxy_mod option for dependency injection during a checkout

fishcakez:
Closes #6 as there won't be a built in sandbox if we use this approach.


josevalim:
:shipit:

PR #11, opened 2016-01-06
Add :owner to DBConnection.Connection, closes #1

### LINKED ISSUE #1: Add owner feature to connection
See: https://github.com/robconery/moebius/issues/37

If the owner dies, the connection process should exit.

josevalim:


lib/db_connection/connection.ex
```diff
@@ -93,6 +93,11 @@ defmodule DBConnection.Connection do
   def init({mod, opts, mode, info}) do
     queue  = if mode == :sojourn, do: :broker, else: :queue.new()
     broker = if mode == :sojourn, do: info
+
+    if owner = Keyword.get(opts, :owner) do
+      Process.link(owner)
```
fishcakez: Can we monitor instead and store the ref in state? Can't rule callback module trapping exists.

josevalim: Ah yes, forgot about it. In case of crashes, should we just exit with the same reason as in the DOWN message?

fishcakez: Yes lets do that.

fishcakez: Or just stop with `:shutdown` might be fine here.

josevalim:
Updated.

lib/db_connection/connection.ex
```diff
@@ -93,8 +93,12 @@ defmodule DBConnection.Connection do
   def init({mod, opts, mode, info}) do
     queue  = if mode == :sojourn, do: :broker, else: :queue.new()
     broker = if mode == :sojourn, do: info
+
+    owner     = Keyword.get(opts, :owner)
+    owner_ref = owner && Process.monitor(owner)
```
fishcakez: we should check `mode == :connection` so poolled workers dont get an owner

josevalim:
Updated.

PR #14, opened 2016-01-09
Return broker pid from sojourn's start_link

fishcakez:
@josevalim thoughts?


josevalim:
It looks good. I will very likely need to re-use both the supervisor and the watcher, so I can make them generic later on. :)

PR #15, opened 2016-01-09
DBConnection.Ownership

josevalim:
* [x] Initial working implementation with passing tests
* [x] Support ownership mode
* [x] Use supervisors
* [x] Cache with ETS
* [x] Docs
* [x] Add ownership specific tests


fishcakez:
Shall we merge this and we can work on the remaining tasks in separate PRs or would you like to continue with this branch?

fishcakez:
Checkin/stop/disconnect should be casts to pids so they dot crash if the pool is gone.

We starting the owner process we should link to the manager until the reply is sent. This ensures the manager goes down if a reply is not sent. When sending a reply from another process it should always be linked directly or indirectly via the supervision tree so that the target of the call exits if a reply isn't sent.

josevalim:
@fishcakez I have made checkin async but ran into issues for disconnect and stop. Can you please give it a try? Everything else is done. :bowtie:

fishcakez:
:heart_eyes_cat: great stuff!! I will tag 0.2 in the morning.

josevalim:
Thanks for the remaining fixes! Beautiful!

PR #16, opened 2016-01-10
Introduce execute_many/3

fishcakez:
Adds an efficient method of executing multiple queries inside a transaction.

* Automatically adds a transaction around the call (disabled with `transaction: false`)
* Executes many queries (pipelined in postgrex's case for 1 round trip)
* Stops executing on first error and just returns the error
* If inside a transaction rolls back the transaction on error
* All decoding delayed until after the transaction is committed (potentially after checking in the connection)
* All logging delayed until after the transaction (potentially after checking in the connection)

This means we can execute any number of queries inside a transaction and not block the connection at all to log or decode. Also in postgrex it would only take 3 round trips, however many queries are executed, when wrapping with a transaction.

The caveat is that queries must be prepared beforehand.

Potential issues:

* Ignores results of successful queries if there is an error
* decode_mapper breaks down as its set for all queries, should the API be extended to set per query options? Would that make this too complicated to use?
* Do we want to commit a transaction we cant decode?

### DIFF
--- a/integration_test/cases/execute_many_test.exs
+++ b/integration_test/cases/execute_many_test.exs
@@ -1,0 +1,311 @@
+defmodule ExecuteManyTest do
+  use ExUnit.Case, async: true
+
+  alias TestPool, as: P
+  alias TestAgent, as: A
+  alias TestQuery, as: Q
+  alias TestResult, as: R
+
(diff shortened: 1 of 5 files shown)


fishcakez:
There are no plans to use this at present.

PR #18, opened 2016-01-13
Do a two step checkout to avoid checking out connections before use

josevalim:


lib/db_connection/ownership/owner.ex
```diff
@@ -92,6 +77,24 @@ defmodule DBConnection.Ownership.Owner do
...
   end
 
+  def handle_call(:init, _from, state) do
+    %{pool: pool, pool_mod: pool_mod, pool_opts: pool_opts} = state
+
+    try do
+      pool_mod.checkout(pool, pool_opts)
+    catch
+      kind, reason ->
+        {:stop, {:shutdown, "no checkout"}, {kind, reason, System.stacktrace}, state}
```
fishcakez: Lets reply with `:error` and re-raise the exception. Should not reraise an exception from another process as the stacktrace is impossible and it is confusing.

fishcakez:
I think this is the right fix. I'll merge so you dont have to rebase :D.

PR #20, opened 2016-01-23
Remove :backoff dependency

### LINKED ISSUE #19: Move backoff into the repository
I am considering if we should move backoff to the repository as db_connection_backoff.erl for two reasons:

1. Solve it in a way that's compatible with old and new Erlang versions alike
2. Remove one dependency

Thoughts?

fishcakez:
Closes #19.

The backoff API is made slightly clearer (what is `:normal`?), a new backoff type has been added (`:rand`) and there are TODOs to remove backwards compatible code in 1.0.

However perhaps we should move this to Connection as it is likely to be useful there?

### DIFF
--- a/integration_test/cases/backoff_test.exs
+++ b/integration_test/cases/backoff_test.exs
@@ -17,8 +17,7 @@ defmodule BackoffTest do
       end]
     {:ok, agent} = A.start_link(stack)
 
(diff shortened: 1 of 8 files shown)


fishcakez:
Wow that test failure, maybe we should remove `+T 9`

josevalim:
:shipit:

PR #21, opened 2016-01-29
Add shared ownership mode

josevalim:


fishcakez:
Do we need shared and automatic? We could run tests with shared.

lib/db_connection/ownership/manager.ex
```diff
@@ -77,11 +80,25 @@ defmodule DBConnection.Ownership.Manager do
...
+        {:reply, :already_shared, state}
+      true ->
+        share_and_reply(state, pid)
+    end
+  end
+  def handle_call({:mode, {:shared, pid}}, _from, state) do
+    share_and_reply(state, pid)
+  end
+  def handle_call({:mode, mode}, _from, %{mode_ref: ref} = state) do
+    ref && Process.demonitor(ref, [:flush])
```
fishcakez: I don't think we want to demonitor here because owner must still be in checkouts.

josevalim: Yes, I refactored the code and forgot this, good catch.

josevalim:
> Do we need shared and automatic? We could run tests with shared.

We need automatic because it allows multiple processes to run with different connections.

fishcakez:
Could we get a test to go with that last commit? I think something like changing the mode explicitly from shared to automatic we correctly handle the exit of the shared pid and can get an automatic checkout afterwards. That is a complicated case we could easily regress on. Otherwise I think we are good to ship this.

josevalim:
I tried to but failed. Let me give it another ago. :D

josevalim:
The reason I am getting stuck in the test is because we are waiting for a process to exit and there is a race condition as the manager may not have processed the DOWN message in time. So we were supposed to be able to checkout but we cannot because the manager still didn't receive the DOWN message. :(

fishcakez:
But the mode has been changed to automatic at that point (before shared exit), so we should be able to automatically start an owner that blocks until the connection owned by the shared process checks in?

fishcakez:
Oh wait I see. Ideally we want to checkout from the manager after the DOWN is handled.

josevalim:
@fishcakez pushed a test now. The tracing tip was great!

josevalim:
@fishcakez let me know if you approve this PR and I will cut a new release :D

fishcakez:
:shipit:

PR #25, opened 2016-02-18
Fix typo in db_connection.ex

Theemuts:
Hi, 

I thought it should be GenServer rather than GenSever.


fishcakez:
Thanks @Theemuts, could you squash to one commit please?

PR #34, opened 2016-04-26
Improve exception error messages

fishcakez:
@josevalim can you review error messages? I have reduced the information for sandbox timeout messages but should be enough information.

### DIFF
--- a/integration_test/sojourn/test_helper.exs
+++ b/integration_test/sojourn/test_helper.exs
@@ -1,4 +1,7 @@
-ExUnit.start([capture_log: :true, assert_receive_timeout: 500])
+ExUnit.start([capture_log: :true, assert_receive_timeout: 500,
+              exclude: [:queue_disconnected]])
+
+
 
 Code.require_file "../../test/test_support.exs", __DIR__
 
--- a/lib/db_connection/connection.ex
+++ b/lib/db_connection/connection.ex
@@ -55,15 +55,15 @@ defmodule DBConnection.Connection do
   end
 
   @doc false
-  def stop({pid, ref}, reason, state, _) do
-    Connection.cast(pid, {:stop, ref, reason, state})
+  def stop({pid, ref}, client, reason, state, _) do
+    Connection.cast(pid, {:stop, ref, client, reason, state})
   end
 
   @doc false
-  def sync_stop({pid, ref}, reason, state, opts) do
+  def sync_stop({pid, ref}, client, reason, state, opts) do
     timeout = Keyword.get(opts, :pool_timeout, @pool_timeout)
     {_, mref} = spawn_monitor(fn() ->
-      sync_stop(pid, ref, reason, state, timeout)
+      sync_stop(pid, ref, client, reason, state, timeout)
     end)
     # The reason is not important as long as the process exited
     # before trying to checkin
@@ -151,7 +151,7 @@ defmodule DBConnection.Connection do
     s = %{s | state: nil, client: :closed, timer: nil, queue: queue}
     case client do
       _ when backoff == :nil ->
-        {:stop, {:shutdown, :disconnect}, s}
+        {:stop, {:shutdown, err}, s}
       {_, :after_connect} ->
         {timeout, backoff} = Backoff.backoff(backoff)
         {:backoff, timeout, %{s | backoff: backoff}}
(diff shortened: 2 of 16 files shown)


fishcakez:
NB: This includes a breaking change for the pool API

integration_test/sojourn/test_helper.exs
```diff
@@ -1,4 +1,7 @@
-ExUnit.start([capture_log: :true, assert_receive_timeout: 500])
+ExUnit.start([capture_log: :true, assert_receive_timeout: 500,
+              exclude: [:queue_disconnected]])
+
+
```
josevalim: Extra new lines? :P

lib/db_connection/connection.ex
```diff
@@ -169,23 +169,25 @@ defmodule DBConnection.Connection do
...
+        message = "connection not available because of disconnection"
+        err = DBConnection.Error.exception(message)
         {:reply, {:error, err}, s}
       %{queue: queue} when queue? == true ->
         client = {ref, Process.monitor(pid)}
         queue = :queue.in({client, timeout, from}, queue)
         {:noreply, %{s | queue: queue}}
       _ when queue? == false ->
-        err = DBConnection.Error.exception("connection not available")
+        message = "connection not available because of queue"
```
josevalim: "because of queue"?

lib/db_connection/connection.ex
```diff
@@ -292,9 +301,11 @@ defmodule DBConnection.Connection do
     end
   end
 
-  def handle_info({:timeout, timer, __MODULE__}, %{timer: timer} = s)
-  when is_reference(timer) do
-    exception = DBConnection.Error.exception("client timeout")
+  def handle_info({:timeout, timer, {__MODULE__, pid, timeout}},
+  %{timer: timer} = s) when is_reference(timer) do
+    message = "client #{inspect pid} timed out because " <>
+    "it checked out the connection for longer than #{timeout}ms"
```
josevalim: Beautiful. :heart:

josevalim:
This looks great. The only one unclear to me is the "because of queue", everything else looks excellent!

lib/db_connection/connection.ex
fishcakez: "connection not immediately available"?

integration_test/sojourn/test_helper.exs
fishcakez: Thanks

fishcakez:
Ah I see there is a race condition in the tests.

Also I think we can remove the backwards incompatible change by creating
the exception in the client process and sending the exception as the
reason. This should smooth things over with the ecto sandbox.

fishcakez:
merged in master

lib/db_connection/connection.ex
ericmj: "connection already checked out" seems apt. Or maybe "connection already checked out and queuing disabled",

fishcakez: It is awkward to say "connection already checked out" because we want to return this error from the pools too, where there can be many connections and none are available.

ericmj: Change to plural in that case maybe? "all connections already checked out"

fishcakez: Unfortunately it may not be true. There is an race condition when using poolboy where poolboy detects a client process has exited and checks out the connection to a new client and the new client tries a checkout before the connection has realised the old client has exited.

ericmj: Do you recommend not using poolboy in that case?

fishcakez: This can only occur when `queue: false` is used. Otherwise the connection will queue the request until the `:DOWN` arrives and then receive an disconnected error.

Poolboy is simpler than sbroker so I would probably recommend poolboy unless requiring the queuing features (max queue length, codel management or rescue-able timeout) provided by sbroker. I will try to add a comparison of the pools to the README today.

josevalim: A good error message needs to have enough information so developers can make a decision of if and how they should act on it. "because of queue" does not give me any information to say if it is a configuration error, a temporary error or something I should investigate. So maybe I would say "connection not available and queueing is disabled".

fishcakez: OK lets go for "connection not immediately available and queuing is disabled" to make it very clear what happened. Likely we can improve things for sbroker's queue dropping exception.

fishcakez: https://github.com/fishcakez/db_connection/commit/81a08526ad61a0f5d4409e32fa2068bbe66f37aa

PR #41, opened 2016-06-05
Update to latest sbroker

fishcakez:
!!WORK IN PROGRESS!!

* Support pool resizing via an sregulator
* Support short circuiting requests via sprotector (on by default)
* Set SASL :alarm_handler alarms when appropriate (on by default, though overload alarm off if using protector)

Need to test defaults, certainly I think `:idle_target` might be wrong, perhaps 5%-10% of `:idle_interval` is better. If `:idle_target` is a higher proportion of `:idle_interval` it will cause the pool to grow quicker and drop slower.

Also I haven't added the ability to load balance brokers.

sbroker master requires OTP 18(/Elixir 1.2), it was costing an unreasonable about of code (500loc) to support a version that wouldn't work well (fallback to :erlang.now/0 for time)


josevalim:

> sbroker master requires OTP 18(/Elixir 1.2), it was costing an unreasonable about of code (500loc) to support a version that wouldn't work well (fallback to :erlang.now/0 for time)

:tada:

PR #43, opened 2016-06-11
Improve error messages and rename Owner to Proxy

josevalim:


lib/db_connection/ownership/owner.ex
```diff
@@ -65,13 +65,21 @@ defmodule DBConnection.Ownership.Owner do
...
+  def handle_info({:DOWN, mon, _, pid, reason},
+                  %{client: {_, _, mon}} = state) do
+    message = "client #{inspect pid} exited with: " <> Exception.format_exit(reason)
     disconnect(message, state)
   end
 
-  def handle_info({:DOWN, ref, _, pid, reason}, %{owner_ref: ref} = state) do
-    message = "owner #{inspect pid} exited: " <> Exception.format_exit(reason)
+  def handle_info({:DOWN, ref, _, pid, reason},
+                  %{owner_ref: ref, client: nil} = state) do
```
fishcakez: since we detect no client here can prevent duplicate logic in `down/2` and just do the checkin instead of creating a message only not to use it?

lib/db_connection/ownership/owner.ex
```diff
@@ -65,13 +65,21 @@ defmodule DBConnection.Ownership.Owner do
...
+    message = "owner #{inspect pid} exited with: " <> Exception.format_exit(reason)
+    down(message, state)
+  end
+
+  def handle_info({:DOWN, ref, _, pid, reason},
+                  %{owner_ref: ref, client: {client, _, _}} = state) do
+    message = "owner #{inspect pid} exited while client #{inspect client} is still running with: " <> Exception.format_exit(reason)
```
fishcakez: this can go straight to `disconnect/2` then

lib/db_connection/ownership/owner.ex
josevalim: We would duplicate the logic though, won't we? Which one do you prefer? :)

josevalim: Also, don't we always use the message, since we use it as the shutdown reason?

fishcakez: ah yes lets keep this

fishcakez: :shipit:

PR #46, opened 2016-06-20
Add ensure_all_started callback to pools

josevalim:


lib/db_connection/pool.ex
```diff
@@ -5,6 +5,12 @@ defmodule DBConnection.Pool do
   """
 
   @doc """
+  Ensure all applications necessary to run the pool are started.
+  """
+  @callback ensure_all_started(opts :: Keyword.t, type :: atom) ::
```
fishcakez: can we use :temporary etc here?

fishcakez:
That failing test is popping all time but it is a race condition in the test :(.

PR #47, opened 2016-06-23
Only send checkout reply once in ownership pool

fishcakez:


fishcakez:
Note the test failure is related to `:sbroker`, ill look at that tomorrow.

josevalim:
:shipit:

PR #51, opened 2016-07-15
Handle exceptions in Query/Connection callbacks in client process

fishcakez:
This adds some major changes in error handling so requires tests plus review. I plan to add a commit which does more error handling in `prepare_execute` (closes prepared query if describe or encode raises).

### DIFF
--- a/integration_test/cases/execute_test.exs
+++ b/integration_test/cases/execute_test.exs
@@ -123,6 +123,92 @@ defmodule ExecuteTest do
       handle_execute: [%Q{}, [:param], _, :state]] = A.record(agent)
   end
 
+  test "execute logs raise" do
+    stack = [
+      fn(opts) ->
+        Process.link(opts[:parent])
+        {:ok, :state}
+      end,
+      fn(_, _, _, _) ->
+        raise "oops"
+      end,
+      {:ok, :state2}
+      ]
+    {:ok, agent} = A.start_link(stack)
+
+    parent = self()
+    opts = [agent: agent, parent: parent]
+    _ = Process.flag(:trap_exit, true)
+    {:ok, pool} = P.start_link(opts)
+    log = fn(entry) ->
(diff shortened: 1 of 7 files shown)


fishcakez:
Note for future: this will require handling result of `{kind, reason, stack}` in Ecto.LogEntry

josevalim:
It looks good to me but we need to be careful with not breaking current Ecto versions as soon as we release this.

fishcakez:
@josevalim: Shall we solve this by rewriting `{kind, class, stack}` as `{:error, %DBConnection.ConnectionError{}}` when logging for now?

I have added the close on raising in describe/encode - we need this for https://github.com/elixir-ecto/postgrex/issues/219.

josevalim:
@fishcakez that would probably be better, yes.

fishcakez:
@josevalim ok tests have been added.

With `prepare_execute` we now close the query on describe exception. Should we do the same for `prepare`? I think we have to for consistency and because the query did not finish preparing . This means a slower checkin if just doing a prepare but that situation is rare - usually inside a transaction.

integration_test/cases/execute_test.exs
```diff
@@ -196,7 +282,7 @@ defmodule ExecuteTest do
 
     assert [
       {:connect, _},
-      {:handle_execute, [%Q{}, [:param], _, :state]}| _] = A.record(agent)
+      {:handle_execute, [%Q{}, [:param], _, :state]} | _] = A.record(agent)
```
fishcakez: oops we fix this mistake later

josevalim:
> With prepare_execute we now close the query on describe exception. Should we do the same for prepare? I think we have to for consistency and because the query did not finish preparing . This means a slower checkin if just doing a prepare but that situation is rare - usually inside a transaction.

Sounds like the way to go indeed.

PR #64, opened 2016-10-14
Update docs and default opts to []

### LINKED ISSUE #63: Incorrect example in docs for DBConnection.prepare
Hi,
I was trying to follow docs (I was implementing an `after_connect` hook) and I noticed example is incorrect:

([db_connection.ex:400](https://github.com/elixir-ecto/db_connection/blob/5cc81fa162add46fc31be47035e69b88330f7872/lib/db_connection.ex#L400:L405))

```elixir
{ok, query}   = DBConnection.prepare(pid, "SELECT id FROM table")
{:ok, result} = DBConnection.execute(pid, query, [])
:ok           = DBConnection.close(pid, query)
```

1. the second argument to `DBConnection.prepare` must be something that implements DBConnection.Query (e.g. `Postgrex.Query`), not a string
2. there's no DBConnection.execute/3, it should be: `DBConnection.execute(pid, query, [], [])`
3. (nitpick) the variable `pid` could be called `conn` to match function head

I'm happy to open up a PR with docs fixes (for some or all points) - what would be the best example of a `query` argument to `prepare/2`? Not sure it makes to assume `Postgrex.Query`, but at the same time it would be good to be concrete.

wojtekmach:
Closes #63

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -422,7 +422,7 @@ defmodule DBConnection do
   See `prepare/3`.
   """
   @spec prepare!(conn, query, opts :: Keyword.t) :: query
-  def prepare!(conn, query, opts) do
+  def prepare!(conn, query, opts \\ []) do
     case prepare(conn, query, opts) do
       {:ok, result} -> result
       {:error, err} -> raise err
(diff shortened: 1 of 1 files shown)


lib/db_connection.ex
```diff
@@ -451,9 +453,10 @@ defmodule DBConnection do
 
   ### Example
 
-      {ok, query, result} = DBConnection.prepare_execute(pid, "SELECT id FROM table WHERE id=$1", [1])
-      {:ok, result2}      = DBConnection.execute(pid, query, [2])
-      :ok                 = DBConnection.close(pid, query)
+      query                = %Query{statement: "SELECT id FROM table WHERE id=$1"}
+      {:ok, query, result} = DBConnection.prepare_execute(conn, [1])
```
fishcakez: Missing query arg

wojtekmach: oops, fixed!

fishcakez:
Thank you :heart:

PR #67, opened 2016-11-10
Add manual ping

fishcakez:
Required a bit more code than I would like because had connection process callback return but it is simple and clear, and most code is tests. Will work with mariaex and postgrex without changes there. For https://github.com/elixir-ecto/ecto/issues/1789.

### DIFF
--- a/integration_test/cases/idle_test.exs
+++ b/integration_test/cases/idle_test.exs
@@ -44,6 +44,7 @@ defmodule TestIdle do
       ping: [:state]] = A.record(agent)
   end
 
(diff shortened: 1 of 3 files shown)


fishcakez:
Ah this will raise if the connection isn't connected. Will need to write a special checkout. This can be unreliable, well a `:pang` might be returned if the previous request timed out even if connection would be successful. Also need to think about what to do if inside a rolling back transaction.

josevalim:
@fishcakez it is also supposedly OK to handle failures in a task and handling exits, no?

PR #70, opened 2016-12-28
Add :caller to ownership error message

### LINKED ISSUE #57: Add `caller` checkout option to list of solutions in `DBConnection.OwnershipError`
When doing similar async tasks to parallel preloads in Ecto it would be useful to use the caller option. We may need to check/test that caller works if a child task uses parallel preloads in Ecto.

/cc @josevalim @nathany

fishcakez:
Closes #57.


fishcakez:
Failed test fixed in https://github.com/elixir-ecto/db_connection/commit/f8d127e493e037cfffb767ffd8510eec4d66c9ed.

PR #72, opened 2017-01-04
Remove unused module attribute

michalmuskala:
I'm not sure it's not a bug this is unused, but I'm not familiar enough with the code to decide.

### DIFF
--- a/lib/db_connection/sojourn.ex
+++ b/lib/db_connection/sojourn.ex
@@ -35,9 +35,8 @@ defmodule DBConnection.Sojourn do
 
   @behaviour  DBConnection.Pool
 
-  @broker    DBConnection.Sojourn.Broker
-  @regulator DBConnection.Sojourn.Regulator
-  @timeout   15_000
+  @broker DBConnection.Sojourn.Broker
+  @timeout 15_000
 
   import Supervisor.Spec
 


lib/db_connection/sojourn.ex
```diff
@@ -35,9 +35,8 @@ defmodule DBConnection.Sojourn do
 
   @behaviour  DBConnection.Pool
 
-  @broker    DBConnection.Sojourn.Broker
-  @regulator DBConnection.Sojourn.Regulator
-  @timeout   15_000
+  @broker  DBConnection.Sojourn.Broker
```
lexmag: I think we should stop making values aligned, it breaks changes history, and makes unnecessary noise in PRs. Looks like `@behaviour` definition is already "broken" because of this reason, it has 2 spaces. :bowtie:

fishcakez:
The attribute isn't used, it was added in the following commit, the correct place is shown on this line: https://github.com/elixir-ecto/db_connection/commit/19640572d5d383f09da3aa22b59e9aa0df7e534d#diff-0586e38a61a4d8ccf461392d83d037bfR4.

lib/db_connection/sojourn.ex
michalmuskala: Yeah, I'm probably guilty of over-aligning things everywhere.

michalmuskala:
The change was to remove the alignment of attributes.

PR #79, opened 2017-03-09
Fix ownership caller to use caller on auto/shared

fishcakez:
Always use :caller when option is set, and fail if :caller does not have a proxy.

We should release patch to fix this issue. A caller process will get a new proxy in `:auto` mode regardless of `caller` and will get the proxy of `owner` in `{:shared, owner}` even if `caller` is not `owner`.

### DIFF
--- a/integration_test/ownership/manager_test.exs
+++ b/integration_test/ownership/manager_test.exs
@@ -111,8 +111,22 @@ defmodule ManagerTest do
       assert_checked_out pool, [caller: parent]
       send parent, :checkin
     end
-
     assert_receive :checkin
+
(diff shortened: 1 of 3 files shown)


josevalim:
> A caller process will get a new process in :auto mode regardless of caller and will get the proxy of owner in {:shared, owner} even if caller is not owner.

What does it mean in practice? Some process may get a connection outside of the specified sandbox?

fishcakez:
> What does it mean in practice? Some process may get a connection outside of the specified sandbox?

Does the edit to the top comment clear this up?

The easiest example is to use parallel preload in Ecto with `:auto` mode. The preload tasks will get their own sandbox. This requires 2 or more preloads.

PR #86, opened 2017-07-03
No longer raise if logging raises

fishcakez:
First commit changes the behaviour and the second commit does a substantial refactor to help with #82 and any future metrics work. Hopefully this will pave the way for future integration with any standard instrumentation lib. This PR uses a stateful approach because it fits well with multiple timestampings. However all "connection_time" work is carried out inside a closure except begin/commit/rollback.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -72,6 +72,8 @@ defmodule DBConnection do
   `run/3` or `transaction/3` fun and using the run/transaction
   connection reference (`t`).
   """
+  require Logger
+
   defstruct [:pool_mod, :pool_ref, :conn_mod, :conn_ref]
 
   @typedoc """
@@ -1189,7 +1191,13 @@ defmodule DBConnection do
 
   defp log(call, query, params, log, times, result) do
     entry = DBConnection.LogEntry.new(call, query, params, times, entry_result(result))
-    log(log, entry)
+    try do
+      log(log, entry)
+    catch
+      kind, reason ->
+        stack = System.stacktrace()
+        log_failed(entry, kind, reason, stack)
+    end
     log_result(result)
   end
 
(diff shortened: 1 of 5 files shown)


lib/db_connection.ex
```diff
@@ -904,45 +922,61 @@ defmodule DBConnection do
...
-  defp checkout(pool, opts) do
+  defp checkout(pool, meter, opts) do
     pool_mod = Keyword.get(opts, :pool, DBConnection.Connection)
-    case apply(pool_mod, :checkout, [pool, opts]) do
+    try do
+      apply(pool_mod, :checkout, [pool, opts])
+    catch
+      kind, reason ->
+        stack = System.stacktrace()
+        {kind, reason, stack, meter}
```
fishcakez: Now long poolboy time outs

lib/db_connection.ex
```diff
@@ -953,243 +987,235 @@ defmodule DBConnection do
...
-  defp transaction_call(:handle_commit), do: :commit
-  defp transaction_call(:handle_rollback), do: :rollback
+  defp run(%DBConnection{} = conn, fun, arg, meter, opts) do
+    fun.(conn, get_info(conn), arg, meter, opts)
+  end
+  defp run(pool, fun, arg, meter, opts) do
+    meter = event(meter, :checkout)
+    with {:ok, conn, conn_info, meter} <- checkout(pool, meter, opts) do
+      try do
+        fun.(conn, conn_info, arg, meter, opts)
```
fishcakez: All "connection time" runs inside this fun (or similar, above/below), which could be instrumented using fun based instrumentation.

lib/db_connection.ex
```diff
@@ -1208,275 +1236,192 @@ defmodule DBConnection do
...
-  defp conclude_meter(conn, conn_state, nil, callback, opts, result) do
-    case handle(conn, conn_state, callback, opts, :idle) do
-      {:ok, _} ->
-        {result, nil}
-      error ->
-        {error, nil}
-    end
+  defp run_begin(conn, conn_info, meter, opts) do
+    meter = event(meter, :begin)
+    handle_trans(conn, conn_info, :handle_begin, meter, opts, :transaction)
```
fishcakez: Transaction calls aren't wrapped in `run` calls with funs but could be as parameters align.

lib/db_connection.ex
```diff
@@ -953,243 +987,235 @@ defmodule DBConnection do
...
+      nil -> nil
+      log -> {log, []}
+    end
   end
 
-  defp log(call, query, params, log, times, result) do
+  defp event(nil, _),
+    do: nil
+  defp event({log, events}, event),
+    do: {log, [{event, :erlang.monotonic_time()} | events]}
```
fishcakez: simple stack based/pure implementation so could wrap any stateful instrumenter

PR #87, opened 2017-07-04
Switch to non-closure based transactions

fishcakez:
Requiring a fun for a transaction and non-local return for a rollback
is a very limiting API. With this breaking change the API becomes more
flexible, in particular it allows transaction to continue over multiple
OTP behaviour callbacks and use of (potentially nested) savepoints.

Removes:
* DBConnection.transaction/3
* DBConnection.rollback/2

Adds:
* DBConnection.checkout/2
* DBConnection.checkin/2
* DBConnection.begin/2
* DBConnection.commit/2
* DBConnection.rollback/2
* DBConnection.checkout_begin/2
* DBConnection.commit_checkin/2
* DBConnection.rollback_checkin/2

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -69,15 +69,15 @@ defmodule DBConnection do
   By default the `DBConnection` provides a single connection. However
   the `:pool` option can be set to use a pool of connections. If a
   pool is used the module must be passed as an option - unless inside a
-  `run/3` or `transaction/3` fun and using the run/transaction
-  connection reference (`t`).
+  `run/3` fun or checked out with `checkout/2` or `checkout_begin/2`
+  and using the connection reference (`t`).
   """
   require Logger
 
   defstruct [:pool_mod, :pool_ref, :conn_mod, :conn_ref]
 
   @typedoc """
-  Run or transaction connection reference.
+  Locked connection reference.
   """
   @type t :: %__MODULE__{pool_mod: module,
                          pool_ref: any,
@@ -508,8 +508,8 @@ defmodule DBConnection do
     * `:queue` - Whether to block waiting in an internal queue for the
     connection's state (boolean, default: `true`)
     * `:timeout` - The maximum time that the caller is allowed the
-    to hold the connection's state (ignored when using a run/transaction
-    connection, default: `15_000`)
+    to hold the connection's state (ignored when using a locked connection,
+    default: `15_000`)
     * `:log` - A function to log information about a call, either
     a 1-arity fun, `{module, function, args}` with `DBConnection.LogEntry.t`
     prepended to `args` or `nil`. See `DBConnection.LogEntry` (default: `nil`)
@@ -564,8 +564,8 @@ defmodule DBConnection do
     * `:queue` - Whether to block waiting in an internal queue for the
     connection's state (boolean, default: `true`)
     * `:timeout` - The maximum time that the caller is allowed the
-    to hold the connection's state (ignored when using a run/transaction
-    connection, default: `15_000`)
+    to hold the connection's state (ignored when using a locked connection,
+    default: `15_000`)
     * `:log` - A function to log information about a call, either
     a 1-arity fun, `{module, function, args}` with `DBConnection.LogEntry.t`
     prepended to `args` or `nil`. See `DBConnection.LogEntry` (default: `nil`)
@@ -620,8 +620,8 @@ defmodule DBConnection do
     * `:queue` - Whether to block waiting in an internal queue for the
     connection's state (boolean, default: `true`)
     * `:timeout` - The maximum time that the caller is allowed the
-    to hold the connection's state (ignored when using a run/transaction
-    connection, default: `15_000`)
+    to hold the connection's state (ignored when using a locked connection,
+    default: `15_000`)
     * `:log` - A function to log information about a call, either
     a 1-arity fun, `{module, function, args}` with `DBConnection.LogEntry.t`
     prepended to `args` or `nil`. See `DBConnection.LogEntry` (default: `nil`)
@@ -671,8 +671,8 @@ defmodule DBConnection do
     * `:queue` - Whether to block waiting in an internal queue for the
     connection's state (boolean, default: `true`)
     * `:timeout` - The maximum time that the caller is allowed the
-    to hold the connection's state (ignored when using a run/transaction
-    connection, default: `15_000`)
+    to hold the connection's state (ignored when using a locked connection,
+    default: `15_000`)
     * `:log` - A function to log information about a call, either
     a 1-arity fun, `{module, function, args}` with `DBConnection.LogEntry.t`
     prepended to `args` or `nil`. See `DBConnection.LogEntry` (default: `nil`)
@@ -714,9 +714,7 @@ defmodule DBConnection do
   connection disconnects all future calls using that connection
   reference will fail.
 
-  `run/3` and `transaction/3` can be nested multiple times but a
-  `transaction/3` call inside another `transaction/3` will be treated
-  the same as `run/3`.
+  `run/3` can be nested multiple times.
 
   ### Options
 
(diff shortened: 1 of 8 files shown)


josevalim:
Do we need all 8 new functions? I assume the last three is for convenience?

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
     end
   end
   def run(pool, fun, opts) do
+    conn = checkout!(pool, opts)
+    try do
+      fun.(conn)
+    after
+      checkin(conn, opts)
+    end
+  end
```
josevalim: New line missing?

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
+  @doc """
+  Acquire a lock on a connection.
+
+  Returns `{:ok, conn}` on success or `{;error, exception}` if there is an
```
josevalim: `;error`

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
+  error.
+
+  To use the locked connection call requests with the connection
```
josevalim: I had a hard time understanding this first sentence.

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
+  reference returned. If the connection disconnects all future calls using that
+  connection reference will fail.
```
josevalim: Can we be clearer on the meaning of fail? Is it `raise` or an error tuple?

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
+        :ok
+      {:error, err} ->
+        raise err
+    end
+  end
+
+  @doc """
+  Begin a transaction.
+
+  Return `{:ok, result}` on sucess or `{:error, exception}` if there was an
```
josevalim: What about `:aborted_transaction` and `:already_in_transaction`? What happens if it is invoked multiple times?

lib/db_connection.ex
josevalim: What happens if I call it multiple times from the same process? Error? I get multiple connections? Other?

lib/db_connection.ex
```diff
@@ -738,45 +736,134 @@ defmodule DBConnection do
...
-  Acquire a lock on a connection and run a series of requests inside a
-  transaction. The result of the transaction fun is return inside an `:ok`
-  tuple: `{:ok, result}`.
+  Release a lock on a connection.
 
-  To use the locked connection call the request with the connection
-  reference passed as the single argument to the `fun`. If the
-  connection disconnects all future calls using that connection
-  reference will fail.
+  Returns `:ok` on success, otherwise `{:error, exception}` on error.
```
josevalim: What happens if the connection was already checked in? Or if it was disconnected between the last command and checking in?

lib/db_connection.ex
fishcakez: Thanks.

lib/db_connection.ex
fishcakez: Rephrased

lib/db_connection.ex
fishcakez: Made clearer (and elsewhere)

lib/db_connection.ex
fishcakez: This approach doesn't work nicely with logging, because these requests will be logged when no query is occurring, and if either of those are returned it will crash loggers.

There are also issues with some DBs. PostgreSQL ignores (but warns on) multiple start transactions, whereas MySQL will do an implicit commit and start a fresh transaction. Also PostgreSQL will implicitly turn committing transaction into rolling back transaction in an aborted state. This means providing the safe/ecto style is also very opinionated, and doesn't fit any database behaviour.

josevalim: Ok, so we need a mechanism to get the transaction status. And if we don't need those operations, maybe there is little to no purpose in providing begin,rollback,commit in isolation? Shouldn't we just use the checkout_begin and variants instead? Shouldn't we aim to have the smallest possible API?

michalmuskala: What about using db_connection with databases that don't have transactions. `checkout` makes sense in that case, but `checkout_begin` does not.

fishcakez: I think we can definitely ditch `commit_checkin` and `rollback_checkin` because the semantics are not 100% clear. There is `checkout` and `checkin` for non-transaction dbs.

I think that `begin` on its own can be useful for `mode: :savepoint` - i believe we will need this feature in the sandbox.

lib/db_connection.ex
```diff
@@ -785,65 +894,382 @@ defmodule DBConnection do
...
+
+  Return `{:ok, conn, result}` on success or `{:error, exception}` if there was
+  an error. If an error is returned the connection is released back to the pool.
+
+  The `conn` is a connection reference that can be used as the first argument to
+  other requests (except `checkout/2` and `checkout_begin/2`). If the connection
+  disconnects all future calls using that connection reference will return (or
+  raise if bang(!) variant) a `%DBConnection.ConnectionError{}` error.
+
+  Use `checkin/2`, `commit_checkin/2` or `rollback_checkin/2` to release the
```
michalmuskala: What will happen in case of plain `checkin/2`?

lib/db_connection.ex
```diff
@@ -785,65 +894,382 @@ defmodule DBConnection do
...
+            return
+        end
+      end
+    log(return, :checkout_begin, :begin, nil)
+  end
+
+  @doc """
+  Acquire a lock on a connection and begin a transaction.
+
+  Return `{:ok, conn, result}` on success or raise an exception if there was
```
michalmuskala: The return is `{conn, result}`

lib/db_connection.ex
```diff
@@ -785,65 +894,382 @@ defmodule DBConnection do
...
+        DBConnection.commit_checkin(conn)
+      end
+  """
+  @spec commit_checkin(t, opts :: Keyword.t) ::
+    {:ok, result} | {:error, Exceptiont.t}
+  def commit_checkin(%DBConnection{} = conn, opts) do
+    return =
+      with {:ok, result, meter}
+            <- run(conn, &run_commit/4, meter(opts), opts) do
+        checkin(conn, opts)
```
michalmuskala: Can `checkin/2` fail here?

lib/db_connection.ex
fishcakez: Nothing, but note we are going to have an option to force `rollback` on checkin (inside the pool).

lib/db_connection.ex
fishcakez: Thanks.

lib/db_connection.ex
fishcakez: Define fail? `checkin/2` will not raise, but it may return an `{:error, %DBConnection.Error{}}` if the socket is disconnected. For the socket to be disconnect at this moment the commit must have disconnected, and for that happen the `with` pattern doesn't match and `checkin/2` isn't called. We are removing this function though because it is not easy to understand.

lib/db_connection.ex
```diff
@@ -785,65 +894,382 @@ defmodule DBConnection do
...
-      {:rollback, {:ok, res, kind, reason, stack, meter}} ->
-        log(meter, :transaction, :rollback, nil, {:ok, res})
-        :erlang.raise(kind, reason, stack)
-      {query, {:error, err, meter}} when query != nil ->
-        log(meter, :transaction, query, nil, {:error, err})
+  @spec begin(conn, opts :: Keyword.t) :: {:ok, result} | {:error, Exception.t}
+  def begin(conn, opts \\ []) do
+    conn
+    |> run(&run_begin/4, meter(opts), opts)
+    |> log(:begin, :begin, nil)
```
fishcakez: As this function would never be used with a pool (unless crazy) we could merge it with `checkout_begin`, such that if the pool is used it performs the current `checkout_begin`. The `pool` version implemented here could be emulated by immediately checking in. This would mean `begin(conn, opts) :: {:ok, conn, result}`.

michalmuskala: What about a case where you check out a connection and do multiple transactions using the same connection? I could imagine this could be useful if you know you need to do couple transactions one after another.

fishcakez: That would be `begin(t, opts):: {:ok, t, result}` which is a subset of `begin(conn, opts) :: {:ok, t, result}`. The merging means that if `conn` is a `GenServer.server` (aka a pool) it performs a checkout and a begin and if `conn` is `DBConnection.t` (aka connection reference) it performs a begin. This supports both nested and multiple serial transactions (should a database support it).

michalmuskala: What in the case of non-transactional dbs. Do they also use `begin`?

fishcakez: We have `checkout(pool, opts) :: {:ok, t} ` and `checkin(t) :: :ok`.

michalmuskala: Ah, so it's only `checkout_begin` we're getting rid of. Then :+1:. I understood it in a way that the only think we're going to leave is `begin`.

fishcakez:
Closed in a favour of #88 #89 #90

PR #88, opened 2017-07-09
Introduce begin, rollback and commit

fishcakez:
* Add begin(!)/1,2 to checkout and begin
* Add rollback(!)/1,2 to rollback and checkin
* Add commit(!)/1,2 to commit (or rollback) and checkin

rollback/2 is overloaded and has old behaviour inside transaction/3 for
backwards compatibility. However transaction/3 is deprecated because is
makes assumptions about database status that may not hold.

New return values from handle_begin/2, handle_rollback/2 and
handle_commit/2 allow to signal mistaken database status.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -76,6 +76,17 @@ defmodule DBConnection do
 
   defstruct [:pool_mod, :pool_ref, :conn_mod, :conn_ref, :conn_mode]
 
+  defmodule TransactionError do
+    defexception [:next, :message]
+
+    def exception(:begin),
+    do: %__MODULE__{next: :begin, message: "transaction is not started"}
+    def exception(:commit),
+    do: %__MODULE__{next: :commit, message: "transaction is already started"}
+    def exception(:rollback),
+    do: %__MODULE__{next: :rollback, message: "transaction is aborted"}
+  end
+
   @typedoc """
   Run or transaction connection reference.
   """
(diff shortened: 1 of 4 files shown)


lib/db_connection.ex
```diff
@@ -817,48 +843,216 @@ defmodule DBConnection do
...
       end)
   """
-  @spec rollback(t, reason :: any) :: no_return
-  def rollback(conn, reason) do
-    %DBConnection{conn_ref: conn_ref, conn_mode: mode} = conn
-    case get_info(conn, nil) do
-      {:error, err, _meter} ->
+  @spec rollback(conn, (opts :: Keyword.t) | reason :: term) ::
+    {:ok, result} | {:error, Exception.t}
+  def rollback(conn, opts \\ [])
```
fishcakez: Given the `transaction/3` behaviour is this incorrect?

lib/db_connection.ex
```diff
@@ -1317,12 +1577,46 @@ defmodule DBConnection do
...
+
+  defp transaction_error(conn, query, opts) do
+    run(conn, &run_transaction_error/5, query, meter(opts), opts)
+  end
+
+  defp run_transaction_error(conn, conn_state, query, meter, opts) do
+    meter = event(meter, query)
+    msg = "can not #{query} inside legacy transaction"
+    err = DBConnection.ConnectionError.exception(msg)
+    delete_disconnect(conn, conn_state, err, opts)
```
fishcakez: disconnect or just error? mark as failed transaction?

lib/db_connection.ex
```diff
@@ -146,42 +157,56 @@ defmodule DBConnection do
...
     {:error | :disconnect, Exception.t, new_state :: any}
 
   @doc """
   Handle committing a transaction. Return `{:ok, result, state}` on success and
-  to continue, `{:error, exception, state}` to abort the transaction and
-  continue or `{:disconnect, exception, state}` to abort the transaction
-  and disconnect.
+  to continue, `{:begin, state}` to open transaction before continuing,
+  `{:rollback, state}` to rollback aborted transaction,
+  `{:error, exception, state}` to abort the transaction and continue or
```
michalmuskala: It might be that I'm not that familiar with db_connection, but what does "abort the transaction & continue" mean here?

lib/db_connection.ex
```diff
@@ -146,42 +157,56 @@ defmodule DBConnection do
...
+  `{:disconnect, exception, state}` to abort the transaction and disconnect.
+
+  A callback implementation should only return `:begin` or `:rollback` if it
+  can determine the database's transaction status without side effect.
```
michalmuskala: Maybe "without a roundtrip" would be clearer here?

lib/db_connection.ex
```diff
@@ -817,48 +843,216 @@ defmodule DBConnection do
...
 
-      {:error, :bar} = DBConnection.transaction(conn, fn(conn) ->
-        DBConnection.rollback(conn, :bar)
-        IO.puts "never reaches here!"
+      {:ok, conn, result} = DBConnection.begin(pool)
+      try do
+        DBConnection.execute!(conn, "SELECT * FROM table", [])
+        DBConnection.commit(conn)
+      after
+        DBConnection.rollback(conn)
```
michalmuskala: How does always calling `rollback` work in this case? Should it only be called in case `execute!` raises?

michalmuskala: I guess this is explained in `rollback` with:
> It is possible to issue multiple rollbacks requests without a begin (`begin/2`
> or `checkout_begin/2`). The semantics are left to the callback
> implementation.

But it still seems a bit wired to me

lib/db_connection.ex
```diff
@@ -817,48 +843,216 @@ defmodule DBConnection do
...
+  ### Options
+
+  See module documentation. The pool and connection module may support other
+  options. All options are passed to `handle_rollback/2`.
+
+  See `begin/2`, `commit/2`, and `transaction/3`.
+
+  ### Example
+
+      {:ok, result} = DBConnection.begin(conn)
```
michalmuskala: `begin` returns a 3 tuple, doesn't it?

lib/db_connection.ex
```diff
@@ -980,6 +1192,17 @@ defmodule DBConnection do
...
 
+  defp checkin(%DBConnection{} = conn, fun, meter, opts) do
+    with {:ok, conn_state, meter} <- fetch_info(conn, meter) do
+      return = fun.(conn, conn_state, meter, opts)
+      checkin(conn, opts)
+      return
+    end
+  end
+  defp checkin(pool, fun, meter, opts) do
+    run(pool, fun, meter, opts)
```
michalmuskala: I'm not sure what will happen in this case. Are we going to check out a connection, call `checkin` from the fun and then check the connection back in?

I'm not sure I understand the use case of calling `checkin` with the pool & without an actual connection.

lib/db_connection.ex
michalmuskala: I don't think we need to disconnect, do we?

lib/db_connection.ex
fishcakez: > Should it only be called in case execute! raises?

Yes, this is simplest/safest way to handle a transaction.



Oops i missed removing `checkout_begin` after copy and paste of docs.



> But it still seems a bit wired to me

We are depending on the database's state, not our own, so the callback implementation will use information from the last request when available.

lib/db_connection.ex
fishcakez: yes, nice catch

lib/db_connection.ex
fishcakez: This is for API completeness. Also someone using the ownership pool may wish to commit/rollback their transaction before returning connection to the inner pool. I am happy to remove this option.

lib/db_connection.ex
fishcakez: In this case it probably needs rewording. The error will cause `{:error, exception}` to be returned by `rollback/2`. For example this might occur if the driver can't do a side effect free transaction status check.

lib/db_connection.ex
fishcakez: We don't need but I worry what might happen if anyone hits this path. If we fail the transaction I guess that stop any issue - i.e. force a rollback at end of `transaction/3`.

lib/db_connection.ex
fishcakez: With a roundtrip is fine. We will need to do this for the sandbox.

lib/db_connection.ex
fishcakez: Sorry to clarify this is safe to call because the connection is checked in by commit so rollback won't do anything if it doesn't raise. This is a standard approach and imo should be the encouraged way as it ensures safety with minimal fuss.

fishcakez:
This change is backwards compatible, so the plan is to soft deprecated `transaction/3`.

lib/db_connection.ex
michalmuskala: Ok, sounds great. If it's an optimal approach maybe we should talk about it in the docs?

fishcakez: Okay, i will add it the examples and mention in the docs for begin.

PR #93, opened 2017-08-22
adding default empty options for a transaction

shdblowers:
Getting errors when testing out mssqlex with updated version of DBConnection because of this.

According to docs this default should still be here.


fishcakez:
Yes it should! Thank you :heart:

Issue #94, opened 2017-09-27
DBConnection should gracefully stop DBConnection.Ownership and DBConnection.Sojourn internals pools


sumerman:
As mentioned in https://github.com/elixir-ecto/ecto/issues/2219#issuecomment-329884307

fishcakez:
Closing in favor of #95.

PR #95, opened 2017-09-29
Cleanly shutdown ownership and sojourn pools (#94)

josevalim:
@fishcakez I am not quite sure how to solve this for sbroker. It seems it has an init callback but it doesn't have a terminate one?

### DIFF
--- a/lib/db_connection/app.ex
+++ b/lib/db_connection/app.ex
@@ -6,11 +6,11 @@ defmodule DBConnection.App do
 
   def start(_, _) do
     children = [
-      worker(DBConnection.Watcher, []),
       supervisor(DBConnection.Task, []),
       supervisor(DBConnection.Sojourn.Supervisor, []),
-      supervisor(DBConnection.Ownership.PoolSupervisor, [])
+      supervisor(DBConnection.Ownership.PoolSupervisor, []),
+      worker(DBConnection.Watcher, [])
     ]
-    Supervisor.start_link(children, strategy: :rest_for_one, name: __MODULE__)
+    Supervisor.start_link(children, strategy: :one_for_all, name: __MODULE__)
   end
 end
--- a/lib/db_connection/app.ex
+++ b/lib/db_connection/app.ex
@@ -6,10 +6,11 @@ defmodule DBConnection.App do
 
   def start(_, _) do
     children = [
+      worker(DBConnection.Watcher, []),
       supervisor(DBConnection.Task, []),
       supervisor(DBConnection.Sojourn.Supervisor, []),
       supervisor(DBConnection.Ownership.PoolSupervisor, [])
-   ]
-    Supervisor.start_link(children, [strategy: :one_for_one, name: __MODULE__])
+    ]
+    Supervisor.start_link(children, strategy: :rest_for_one, name: __MODULE__)
   end
 end
--- a/lib/db_connection/watcher.ex
+++ b/lib/db_connection/watcher.ex
@@ -1,29 +1,56 @@
 defmodule DBConnection.Watcher do
   @moduledoc false
+  @name __MODULE__
 
   use GenServer
 
-  def start_link(owner) do
-    GenServer.start_link(__MODULE__, owner)
+  def start_link do
+    GenServer.start_link(__MODULE__, :ok, name: @name)
   end
 
-  def init(owner) do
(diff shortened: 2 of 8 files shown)


fishcakez:
@josevalim the reason for that it is only used as configuration, like a supervisor, and the callback is called on appups for reconfiguration.

We would have to have a process similar to the file server in otp that lives in the db_connection supervision tree and handles the start/stop of the pool for sbroker. TBH I would take any opportunity to deprecate the sojourn pool as it adds complications, such as this. sbroker does not fit well with the traditional pooling supervision layout because it is not designed to be used this way.

josevalim:
@fishcakez ok, so for now let's merge this if you agree and we can revisit what to do regarding sojourn in an eventual 2.0.

fishcakez:
I think we should solve it for both pools as we are still supporting them. I think this will require having a GenServer that links to ownership manager or the sojourn broker. In the case of a crash of the pool (monitored) it should send an exit signal (that is abnormal, perhaps kill to ensure a trapping exit broker goes down) to the manager/broker. In a case of a crash of the manager it should terminate the pool. What do you think?

josevalim:
@fishcakez basically we take the watcher out of the pool supervisor, put them in their own supervisor, and make sure it terminates and send the proper exit signals? That sounds good to me.

fishcakez:
@josevalim yeah

josevalim:
@fishcakez I have implemented a new PR that uses a central DBConnection.Watcher as we discussed.

lib/db_connection/app.ex
```diff
@@ -6,10 +6,11 @@ defmodule DBConnection.App do
...
   def start(_, _) do
     children = [
+      worker(DBConnection.Watcher, []),
       supervisor(DBConnection.Task, []),
       supervisor(DBConnection.Sojourn.Supervisor, []),
       supervisor(DBConnection.Ownership.PoolSupervisor, [])
-   ]
-    Supervisor.start_link(children, [strategy: :one_for_one, name: __MODULE__])
+    ]
+    Supervisor.start_link(children, strategy: :rest_for_one, name: __MODULE__)
```
fishcakez: This supervision tree isn't quite right because the `Watcher` makes calls to the other supervisors. Therefore it depends on those supervisors and so we want the `Watcher` to be started last/shutdown first. Can we put it last and make this a `:one_for_all`. Therefore we would still shutdown all pools if the `Watcher` crashes. None of the supervisors should ever crash anyway because they all have temporary children.

lib/db_connection/watcher.ex
```diff
@@ -1,29 +1,56 @@
...
+      %{^ref => {supervisor, started_pid, started_ref}} ->
+        Process.demonitor(started_ref, [:flush])
+        Supervisor.terminate_child(supervisor, started_pid)
+        {:noreply, {Map.delete(caller_refs, ref), Map.delete(started_refs, started_ref)}}
+      %{} ->
+        case started_refs do
+          %{^ref => {caller_pid, caller_ref}} ->
+            Process.demonitor(caller_ref, [:flush])
+            Process.unlink(caller_pid)
+            Process.exit(caller_pid, :kill)
```
fishcakez: Let's `:kill` all callers in `terminate/2` as well to be sure we don't leave a trapping exit broker standing with no pool behind it. The supervision tree will handle the pools.

lib/db_connection/watcher.ex
```diff
@@ -1,29 +1,56 @@
...
+    case caller_refs do
+      %{^ref => {supervisor, started_pid, started_ref}} ->
+        Process.demonitor(started_ref, [:flush])
+        Supervisor.terminate_child(supervisor, started_pid)
+        {:noreply, {Map.delete(caller_refs, ref), Map.delete(started_refs, started_ref)}}
+      %{} ->
+        case started_refs do
+          %{^ref => {caller_pid, caller_ref}} ->
+            Process.demonitor(caller_ref, [:flush])
+            Process.unlink(caller_pid)
```
fishcakez: Lets skip the `unlink` because we trap exits and handle exit signals, and technically watcher could be killed by its parent supervisor between `unlink` and `kill`.

fishcakez:
I will fixup this branch

josevalim:

Issue #96, opened 2017-10-10
Do not permanently give connections out when starting the sandbox


josevalim:
See discussion here: https://elixirforum.com/t/phoenix-testing-with-ecto-2-sandbox-access-from-processes/9174/9

@fishcakez is this something we can do?

josevalim:
@fishcakez yes, it is something we can do. :) Please see attached PR.

The first commit reclaims connections when we change the mode to shared. The second commit generalizes the logic and reclaims whenever there is a mode change.

Another option is to make the reclaiming explicit and users can do it whenever they change the mode (or the Ecto sandbox could encapsulate it). Although I believe this provides a better behaviour given that the Ecto sandbox already has this:

```elixir
    # If the mode is set to anything but shared, let's
    # automatically checkin the current connection to
    # force it to act according to the chosen mode.
    if mode in [:auto, :manual] do
      checkin(repo, [])
    end
```

josevalim:
Btw, we have Ecto failures when running on DBConnection master, we may need to make this DBConnection 2.0.

josevalim:
We cannot block because sometimes we need to do operations before the test starts, such as running migrations, seeding immutable data to the DB, etc.

fishcakez:
Can we commence blocking after the setup but before the application is started?

josevalim:
@fishcakez I tried to explore this today but it is tricky because the repository starts with the application so we would need a mechanism where we configure the repository before it is even running and also set flags such as `mix test --no-start`. :(

fishcakez:
My previous advice in this situation has been to stop required apps, set shared mode and restart them. I think we need to make it more obvious when something like this is required, and what these processes are. Perhaps we can log a warning including all the processes that currently own a sandbox when the mode changes to shared, or more drastically kill them.

josevalim:
Can you really restart the app? The issue is that we can only set the mode after the app starts and by then the service may already have acquired a connection.

fishcakez:
It depends if the repo is in the app or not. If it is in the same app it requires using configuration to set the mode (or defaulting to shared). It would probably be done using an `init` callback to switch the configuration based on another side effect.

josevalim:
@fishcakez which is a lot of indirection to get the desired behaviour. I can totally understand not merging the proposed PR but we need better alternatives so we solve 1. the discovery issue and 2. the issue in itself.

fishcakez:
[...]

PR #97, opened 2017-10-10
Checkin previous connections when setting ownership mode to shared

josevalim:
This fixes a bug where processes started before the ownership
mode is set to shared holds on to the previous connection,
causing confusion.

See #96


jerel:
I ran into an issue where contexts called from Phoenix channels were not caught by the sandbox (the repo they are using is in another umbrella app). Restarting the web app in the test setup fixes the issue but it seems like the sort of thing that could surprise a few people. Maybe it's time for a patch release?

fishcakez:
This is part of master which is for v2.0 unfortunately. We might be able to backport though.

PR #104, opened 2017-11-26
Allow declare to prepare new query

fishcakez:
Follow up to #91 to allow the same feature for streams/cursor. This will allow us to use prepared queries in Ecto for streams.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -268,13 +268,15 @@ defmodule DBConnection do
   @doc """
   Declare a cursor using a query prepared by `handle_prepare/3`. Return
   `{:ok, cursor, state}` to start a cursor for a stream and continue,
-  `{:error, exception, state}` to return an error and continue or
-  `{:disconnect, exception, state}` to return an error and disconnect.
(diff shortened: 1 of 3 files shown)


lib/db_connection.ex
```diff
@@ -1330,8 +1332,8 @@ defmodule DBConnection do
   @doc """
   Declare a cursor.
 
-  Returns `{:ok, cursor}` on success or `{:error, exception}` if there was
-  an error.
+  Returns `{:ok, cursor}` or `{:ok, query, cursor}` on success or
```
josevalim: Do we have both return types for necessity or backwards compatibility? If backwards compatibility, it would be nice to document it as so.

fishcakez: In postgrex PR both are returned. The new 3-tuple is only returned when the query result types change but we could always return it. The reason to return a new query is because we are able to survive type changes cases if parameter types are castable (from old to new) at the database and the result type can be anything. I think it will make sense for consistency to always return a query and leave the 2-tuple for backwards compatibility. However I'm not sure yet if we should encourage drivers to always return 3-tuple yet because haven't updated mariaex/ecto.

PR #111, opened 2018-01-08
Raise RuntimeError when crashing on connect

tetiana12345678:


integration_test/cases/error_on_connect_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule ErrorOnConnectTest do
```
fishcakez: This test won't run. I think we should rename `backoff_test` to `connect_test` and add this test to that module. Be sure to update `integration_test/tests.exs`.

lib/db_connection/connection.ex
```diff
@@ -131,7 +131,17 @@ defmodule DBConnection.Connection do
     %{mod: mod, opts: opts, backoff: backoff, after_connect: after_connect,
       idle: idle, idle_timeout: idle_timeout, regulator: regulator,
       lock: lock} = s
-    case apply(mod, :connect, [opts]) do
+    try do
+      apply(mod, :connect, [opts])
+    rescue e ->
+      stack = System.stacktrace()
+      msg = """
+      Connect raised a #{inspect e.__struct__} error. The parameters passed in are hidden, as
```
fishcakez: `The exception details are hidden` rather than `parameters`.

PR #115, opened 2018-02-20
Fix documented default in Broker

ericentin:
It's actually 128, as can be verified a bit lower in the file: https://github.com/elixir-ecto/db_connection/compare/master...ericentin:patch-1#diff-5a7d4ad5e69065856a2a26d8253c1ddcL151


fishcakez:
There won't be another release on master with this module, can you rebase for v1.1?

PR #117, opened 2018-03-02
remove pid when logger error for easy aggregate error log

redink:
`pid` is not useful here, but it will not easy to aggregate error log with `pid`.

### DIFF
--- a/lib/db_connection/connection.ex
+++ b/lib/db_connection/connection.ex
@@ -163,7 +163,7 @@ defmodule DBConnection.Connection do
         raise err
       {:error, err} ->
         Logger.error(fn() ->
-          [inspect(mod), ?\s, ?(, inspect(self()), ") failed to connect: " |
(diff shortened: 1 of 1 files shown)


josevalim:
If the purpose is aggregation, then I would suggest you to enable the Logger tags and include the application as a tag. This way you can aggregate everything that comes from DBConnection. We can also add more tags but I don't think we should remove useful information from the message (and I do find the PID useful).

Thank you.

redink:
But how to enable Logger tags ?

I only know how to use `metadata`, and if I want to use `metadata`, I have to change the method to call `Logger.error`. So I am not sure how to do without change this code.

fishcakez:
Mix/Logger will add the metadata `application: :db_connection` to the `Logger.error` calls. Also a callback module can call `Logger.metadata` in the `connect` callback to add extra `metadata`. Note that these calls can also be identified with `module: DBConnection.Connection` metadata.

PR #118, opened 2018-03-06
Bump elixir version to 1.4

m13m:


josevalim:

fishcakez:
Can we revert this please unless there is a reason to bump?

fishcakez:
> Can we revert this please unless there is a reason to bump?

Ah please ignore I thought this was bumping `db_connection` but it is only the examples ♥️

PR #124, opened 2018-06-11
Use optional_callbacks instead of relying on default implementations

josevalim:
This commit also moves the handling of replies to each callback
instead of a main `handle` function as otherwise we would need
to push `function_exported?/3` concerns into it.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -378,145 +378,14 @@ defmodule DBConnection do
   """
   @callback disconnect(err :: Exception.t, state :: any) :: :ok
 
+  @optional_callbacks handle_first: 4, handle_next: 4, handle_info: 2, ping: 1
+
   @doc """
-  Use `DBConnection` to set the behaviour and include default
-  no-op implementations for `ping/1` and `handle_info/2`.
+  Use `DBConnection` to set the behaviour.
   """
   defmacro __using__(_) do
     quote location: :keep do
       @behaviour DBConnection
-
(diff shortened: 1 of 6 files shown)


lib/db_connection.ex
```diff
@@ -378,145 +378,14 @@ defmodule DBConnection do
   """
   @callback disconnect(err :: Exception.t, state :: any) :: :ok
 
+  @optional_callbacks handle_first: 4, handle_next: 4, handle_info: 2, ping: 1
```
fishcakez: * handle_first/next should be removed as they have been replaced by handle_fetch in 2.0, keeping those after a major version bump would be confusing and prevent us simplifying the code
* handle_info wont be called with the new pool so we should remove it
* ping should become required as we no longer have active mode with the new pool, so rely on ping to detect network drops

josevalim: I have done removed handle_first and handle_next. I have also made `ping` required but I am not sure about `handle_info/2`. The TCP example you have in examples defines `handle_info/2`... if that is not going to be called, how can users implement that part of the code?

fishcakez: They can't anymore.

josevalim:
@fishcakez I am struggling to remove handle_info. So I want to merge this as is and then add a note to remove the handle_info callback in another PR. :)

fishcakez:
Ok lets merge this then. Lets remove poolboy and connection before we remove handle_info.

PR #125, opened 2018-06-11
Remove Sojourn pool

josevalim:


josevalim:

michalmuskala:
DBConnection 2 will only have one built-in pool based on ets that was introduced in https://github.com/elixir-ecto/db_connection/pull/108

The new pool is faster and simpler and allows for some additional nice things - for example regular exceptions for timeouts instead of exits from poolboy. In our benchmarks, the new pool is also just 10% slower than not using a pool at all, so we plan to remove the "no pool" option as well. All of this means significant simplifications of the codebase, which is a priority for us at the moment, given we (or more specifically @fishcakez) don't get help with maintenance of it.

PR #128, opened 2018-06-14
Add more logging for ownership errors.

scouten:
This PR attempts to address a challenge with the difficult-to-debug `(DBConnection.ConnectionError) owner #PID<...> exited while client #PID<...> is still running` error. (Which code _started_ that connection? The error message doesn't give sufficient context to answer that question, so the warning is rarely actionable.)

With this PR the stack trace from the code that checks out the connection is retained and added to the error message should there be an unexpected exit.

Consider this more of a rough draft of a PR. I am concerned about the overhead involved and think that it might be a good idea to add an explicit opt-in config option to enable this tracking.

### DIFF
--- a/lib/db_connection/ownership/proxy.ex
+++ b/lib/db_connection/ownership/proxy.ex
@@ -19,13 +19,20 @@ defmodule DBConnection.Ownership.Proxy do
   end
 
   def checkout(proxy, opts) do
+    stack =
+      try do
+        raise "x"
+      rescue
+        _ -> System.stacktrace()
+      end
+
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @scouten! This is a great idea, I like this a lot. I think we can do like GenServer and use `Process.info(pid, :current_stacktrace)` to get the information both of the checkout process and the one that is still running. This way we can do it only for ownership (which is mostly for tests). I am currently reworking the ownership pool so I will make sure to revisit this.

josevalim:
Closing in favor of #135.

PR #130, opened 2018-06-18
Remove ownership pool and always use the connection pool

josevalim:


josevalim:
@fishcakez this is ready for review. :) If you have an idea of what are the next steps after this, please let me know.

PR #135, opened 2018-08-21
Add client information on owner down

josevalim:


lib/db_connection/ownership/proxy.ex
```diff
@@ -20,13 +20,15 @@ defmodule DBConnection.Ownership.Proxy do
   end
 
   def checkout(proxy, opts) do
+    {:current_stacktrace, stack} = Process.info(self(), :current_stacktrace)
```
fishcakez: Can we extend this so the manager gets the stacktrace too? This will allow us to add it the ownership debug logs.

josevalim: I am not sure how we can pass this information to the manager on master. The manager would rather have to "read" this information too, as we now have to do this on the server while the client is waiting for the checkout. So we can treat those as two separate things.

josevalim:
Closing in favor of #164.

PR #139, opened 2018-08-22
Remove begin/rollback/commit explicit API

josevalim:
We can add it back later but for now YAGNI.


josevalim:

PR #141, opened 2018-09-01
Automatically reprepare query on EncodeError

### LINKED ISSUE #123: Prepare and retry some encode failures
If we have a prepared query struct with stale types, and we try execute that query with parameters for the new types the encode will fail without hitting the database. To handle this gracefully we should rescue `DBConnection.EncodeError`exceptions (or similarly unique exception) on encode calls to the query protocol. If it is an execute call we should go to prepare_execute to prepare a new version of the query and try to encode again. Similarly for declare/prepare_declare. For other functions we rescue EncodeError and return it in error tuple.

josevalim:
Closes #123


integration_test/cases/stream_test.exs
```diff
@@ -72,6 +72,48 @@ defmodule StreamTest do
       ] = A.record(agent)
   end
 
+  test "stream reprapres query on encode error" do
```
fishcakez: typo

lib/db_connection.ex
```diff
@@ -1071,9 +1087,11 @@ defmodule DBConnection do
     end
   end
 
-  defp encode(query, params, meter, opts) do
+  defp maybe_encode(query, params, meter, opts) do
     try do
       DBConnection.Query.encode(query, params, opts)
+    rescue
+      DBConnection.EncodeError -> :retry
```
fishcakez: For clarity could we rename `:retry` to be `:prepare`, as it is not a straight retry, we have to do some extra work (prepare). Also we should return the existing `meter`, so its `{:prepare, meter}` so that we create it once.

lib/db_connection.ex
```diff
@@ -80,15 +80,19 @@ defmodule DBConnection do
 
   defstruct [:pool_mod, :pool_ref, :conn_mod, :conn_ref, :conn_mode]
 
+  defmodule EncodeError do
+    defexception [:message]
```
fishcakez: Do we want to support `:value` for the term that was failed to be encoded? For Ecto this is the wrong place to handle this though so don't think we need it.

josevalim:
Btw, this patch does not work for Postgrex as is. If we attempt to prepare
an already prepared query, it will error. So we need to do something,
either by augmenting the query protocol or by passing an option to make
this work.

However, even by doing those changes, this is not enough to fix Ecto tests
on db_connection (there are 2 and one is related to this, you can see on
CI).
-- 


josevalim:
So it seems this patch is wrong? We should not call parse again when
falling back from execute to prepare_execute?
-- 


fishcakez:
Correct don't need to parse when it's already been done.

On 2 Sep 2018 00:30, "José Valim" <notifications@github.com> wrote:

So it seems this patch is wrong? We should not call parse again when
falling back from execute to prepare_execute?

-- 


—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/elixir-ecto/db_connection/pull/141#issuecomment-417911026>,
or mute the thread

josevalim:
I have fixed the typo, changed it so we don't parse twice and moved to `{:prepare, meter}` as return. It should be good to go. I will try it with Postgrex master son.

PR #145, opened 2018-09-07
Check the status on DBConnection.run

### LINKED ISSUE #122: DBConnection.run should fail if the connection state before and after are not the same

josevalim:
Closes #122.

### DIFF
--- a/integration_test/cases/client_test.exs
+++ b/integration_test/cases/client_test.exs
@@ -136,11 +136,11 @@ defmodule ClientTest do
         send(opts[:parent], :reconnected)
         {:ok, :new_state}
       end,
+      {:idle, :new_state},
+      {:idle, :new_state},
       fn(_, _, _, _) ->
         throw(:oops)
       end,
-      {:idle, :new_state},
-      {:idle, :new_state},
       {:ok, %R{}, :newer_state}]
     {:ok, agent} = A.start_link(stack)
 
@@ -152,6 +152,14 @@ defmodule ClientTest do
     try do
       P.run(pool, fn(conn) ->
         assert_receive :reconnected
+
+        spawn_link(fn ->
+          _ = Process.put(:agent, agent)
+          assert P.run(pool, fn(_) -> :result end) == :result
+          send(parent, :done)
+        end)
+
+        assert_receive :done
         P.execute(conn, %Q{}, [:param])
       end, [timeout: 100])
     catch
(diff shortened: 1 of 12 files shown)


integration_test/cases/client_test.exs
```diff
@@ -127,26 +152,30 @@ defmodule ClientTest do
...
         P.execute(conn, %Q{}, [:param])
-      end, [timeout: 0])
+      end, [timeout: 100])
     catch
       :throw, :oops ->
         :ok
     end
 
-    assert_receive :done
+    spawn_link(fn ->
```
fishcakez: Don't need to spawn link here? Though comment above should handle the test and we dont need this.

integration_test/cases/transaction_test.exs
```diff
@@ -965,10 +965,7 @@ defmodule TransactionTest do
 
     assert P.status(pool, opts) == :idle
     assert P.status(pool, opts) == :transaction
-    assert P.run(pool, fn(conn) ->
-      assert P.status(pool, [queue: false] ++ opts) == :error
-      assert P.status(conn, opts)
-    end, opts)
+    assert P.status(pool, [queue: false] ++ opts) == :error
```
fishcakez: Thus doesn't test same thing, this test is that we get an `:error` atom when can't checkout from pool - so that handle_status is not called for this one.

integration_test/cases/client_test.exs
```diff
@@ -127,26 +152,30 @@ defmodule ClientTest do
     try do
       P.run(pool, fn(conn) ->
         assert_receive :reconnected
-        spawn_link(fn() ->
```
fishcakez: We want to spawn_link here so that checkout occurs before `run` or `execute` has chance to fail/checkin. I think this previous code should work with timeout of `100`.

integration_test/cases/idle_test.exs
```diff
@@ -29,12 +29,11 @@ defmodule TestIdle do
     {:ok, agent} = A.start_link(stack)
 
     opts = [agent: agent, parent: self(), idle_timeout: 50, idle_interval: 50]
-    {:ok, pool} = P.start_link(opts)
+    {:ok, _} = P.start_link(opts)
     assert_receive {:hi, conn}
     assert_receive {:pong, ^conn}
     assert_receive {:pong, ^conn}
     send(conn, {:continue, self()})
-    P.run(pool, fn(_) -> :ok end)
```
fishcakez: Are we able to keep this? The test ensures we resume pings after checking back in.

integration_test/cases/client_test.exs
josevalim: It didn't work with a higher timeout either. There is a race on which arguments are popped from the agent.

integration_test/cases/transaction_test.exs
josevalim: I moved this test below: https://github.com/elixir-ecto/db_connection/pull/145/files#diff-c000545f7e7485c4bc7ed2cc55e6f5b8R1021

integration_test/cases/idle_test.exs
josevalim: I don't know why it was removed actually. I will investigate.

josevalim:

PR #148, opened 2018-09-10
Share messages/ETS logic between built in pools

fishcakez:


fishcakez:
Need to improve queue logic in proxy to handle timeouts.

josevalim:
Great job! It is less LOCs too!

PR #154, opened 2018-09-16
Add pre and post callbacks on ownership

josevalim:


josevalim:
We need to remove pdict in order to finish this.

PR #155, opened 2018-09-17
Remove the process dictionary

josevalim:
@fishcakez, please review.

I have dropped comments on the relevant bits. Also please review the changes on holder, task and connection.

### DIFF
--- a/integration_test/cases/client_test.exs
+++ b/integration_test/cases/client_test.exs
@@ -81,10 +81,14 @@ defmodule ClientTest do
       {:handle_status, _}] = A.record(agent)
   end
 
-  test "reconnect when client timeout and then disconnects" do
+  test "reconnect when client timeout and then returns ok even when disconnected" do
     stack = [
       {:ok, :state},
       {:idle, :state},
+      fn(_, _, _, _) ->
+        assert_receive :reconnected
+        {:ok, %Q{}, %R{}, :new_state}
+      end,
       :ok,
       fn(opts) ->
         send(opts[:parent], :reconnected)
@@ -101,9 +105,7 @@ defmodule ClientTest do
     {:ok, pool} = P.start_link(opts)
 
     assert P.run(pool, fn(conn) ->
-      assert_receive :reconnected
-
-      assert {:error, %DBConnection.ConnectionError{message: "connection is closed"}} =
+      assert {:ok, %Q{}, %R{}} =
                P.execute(conn, %Q{}, [:first])
 
       spawn_link(fn() ->
@@ -120,6 +122,56 @@ defmodule ClientTest do
     assert [
       {:connect, _},
       {:handle_status, _},
+      {:handle_execute, [%Q{}, [:first], _, :state]},
+      {:disconnect, _},
+      {:connect, _},
+      {:handle_status, _},
+      {:handle_status, _},
+      {:handle_execute, [%Q{}, [:second], _, :new_state]}] = A.record(agent)
+  end
+
+test "reconnect when client timeout and then returns disconnected when disconnected" do
+    stack = [
+      {:ok, :state},
+      {:idle, :state},
+      fn(_, _, _, _) ->
+        assert_receive :reconnected
+        {:disconnect, DBConnection.ConnectionError.exception("oops"), :new_state}
+      end,
+      :ok,
+      fn(opts) ->
+        send(opts[:parent], :reconnected)
+        {:ok, :new_state}
+      end,
+      {:idle, :new_state},
+      {:idle, :new_state},
+      {:ok, %Q{}, %R{}, :newer_state}]
+    {:ok, agent} = A.start_link(stack)
+
+    parent = self()
+
+    opts = [agent: agent, parent: parent]
+    {:ok, pool} = P.start_link(opts)
+
+    assert P.run(pool, fn(conn) ->
+      assert {:error, %DBConnection.ConnectionError{message: "oops"}} =
+               P.execute(conn, %Q{}, [:first])
+
+      spawn_link(fn() ->
+        _ = Process.put(:agent, agent)
+        assert P.run(pool, fn(_) -> :result end) == :result
+        send(parent, :done)
+      end)
+      :result
+    end, [timeout: 100]) == :result
+
+    assert_receive :done
+    assert P.execute(pool, %Q{}, [:second]) == {:ok, %Q{}, %R{}}
+
+    assert [
+      {:connect, _},
+      {:handle_status, _},
+      {:handle_execute, [%Q{}, [:first], _, :state]},
       {:disconnect, _},
       {:connect, _},
       {:handle_status, _},
@@ -131,7 +183,9 @@ defmodule ClientTest do
     stack = [
       {:ok, :state},
       {:idle, :state},
-      :ok,
+      fn(_, _, _, _) ->
+        throw(:oops)
+      end,
       fn(opts) ->
         send(opts[:parent], :reconnected)
         {:ok, :new_state}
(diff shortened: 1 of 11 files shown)


integration_test/cases/client_test.exs
```diff
@@ -104,7 +102,10 @@ defmodule ClientTest do
 
     assert P.run(pool, fn(conn) ->
       assert_receive :reconnected
-      assert {:error, %RuntimeError{}} = P.execute(conn, %Q{}, [:param])
+
+      assert {:error, %DBConnection.ConnectionError{message: "connection is closed"}} =
+               P.execute(conn, %Q{}, [:first])
```
josevalim: Because the ets entry is removed, this command will never work, so we never get to invoke the agent. So I changed the return result here and removed the :disconnect entry from the agent.

integration_test/cases/client_test.exs
```diff
@@ -78,7 +78,6 @@ defmodule ClientTest do
       {:disconnect, _},
       {:connect, _},
       {:handle_status, _},
-      {:handle_status, _},
```
josevalim: Since we disconnect, the lose the ets table and the stage, so we never get to call handle_status/2 the second time around.

integration_test/cases/client_test.exs
josevalim: I would say this whole test is useless, because the scenario "when client timeout and then disconnects" is no longer possible.

integration_test/cases/client_test.exs
```diff
@@ -138,9 +138,6 @@ defmodule ClientTest do
       end,
       {:idle, :new_state},
       {:idle, :new_state},
-      fn(_, _, _, _) ->
-        throw(:oops)
-      end,
```
josevalim: Same here. We never get to invoke this. So I would say this whole test is useless, because the scenario "when client timeout and then crashes" is no longer possible.

lib/db_connection.ex
```diff
@@ -1192,152 +1167,132 @@ defmodule DBConnection do
...
+    pool_ref
+    |> Holder.handle(:handle_prepare, [query], opts)
+    |> handle_common_result(conn, event(meter, :prepare))
   end
 
-  defp run_prepare_execute(conn, conn_state, query, params, meter, opts) do
-    with {:ok, query, meter}
-          <- run_prepare(conn, conn_state, query, meter, opts),
-         {:ok, params, meter} <- encode(conn, query, params, meter, opts),
-         {:ok, conn_state, meter} <- fetch_info(conn, meter),
```
josevalim: Note we no longer call this `fetch_info`. It is unclear if this `fetch_info` was here to check the status or to get the state. Assuming it was only for the state, then this call is definitely no longer needed.

lib/db_connection/holder.ex
```diff
@@ -178,19 +244,17 @@ defmodule DBConnection.Holder do
     end
   end
 
-  defp done(pool_ref, tag, state, info) do
+  defp done(pool_ref, status, tag, info) do
```
josevalim: Note that `done` now also expects the status to reset the connection to.

lib/db_connection/task.ex
```diff
@@ -15,39 +17,36 @@ defmodule DBConnection.Task do
...
     receive do
-      {:go, ^ref, mon} ->
-        Process.unlink(conn)
-        pool = {:via, __MODULE__, {{conn, mon}, mod, state}}
-        _ = DBConnection.run(pool, make_fun(fun), [holder: __MODULE__] ++ opts)
+      {:"ETS-TRANSFER", holder, ^parent, {:checkin, ref, _extra}} ->
+        Process.unlink(parent)
+        pool_ref = DBConnection.Holder.pool_ref(pool: parent, reference: ref, holder: holder)
+        checkout = {:via, __MODULE__, pool_ref}
+        _ = DBConnection.run(checkout, make_fun(fun), [holder: __MODULE__] ++ opts)
```
josevalim: Note I still kept holder as an option (**only here** and it is not made part of the state). I could move the DBConnection.run/3 logic in here, but it would lead to tons of duplication.

lib/db_connection/holder.ex
```diff
@@ -46,30 +65,79 @@ defmodule DBConnection.Holder do
...
-    holder
+  @spec handle(pool_ref :: any, fun :: atom, args :: [term], Keyword.t) :: tuple
+  def handle(pool_ref, fun, args, opts) do
+    pool_ref(holder: holder) = pool_ref
+
+    try do
+      :ets.lookup(holder, :conn)
+    rescue
+      e in ArgumentError ->
+        {:catch, :error, e, System.stacktrace()}
```
josevalim: This is not ok!

lib/db_connection.ex
```diff
@@ -999,87 +989,61 @@ defmodule DBConnection do
...
         {kind, reason, stack, meter}
 
       other ->
-        bad_return!(other, conn, conn_state, meter)
+        bad_return!(other, conn, meter)
     end
   end
 
-  defp bad_return!(other, conn, conn_state, meter) do
+  @compile {:inline, bad_return!: 3}
```
fishcakez: What stack does this give us in the logs? Does inlining keep the calling function in the stack even when `bad_return!` is called in the tail?

lib/db_connection/holder.ex
```diff
@@ -46,30 +65,106 @@ defmodule DBConnection.Holder do
...
+      [conn(status: {:error, _})] ->
+        msg = "connection is closed"
+        {:disconnect, DBConnection.ConnectionError.exception(msg), _state = :unused}
+
+      [conn(status: :aborted)] when type != :cleanup ->
+        msg = "transaction rolling back"
+        {:disconnect, DBConnection.ConnectionError.exception(msg), _state = :unused}
+
+      [conn(module: module, state: state)] ->
+        try do
```
fishcakez: Can we move this try to a separate function to removing the nesting?

lib/db_connection/holder.ex
```diff
@@ -46,30 +65,106 @@ defmodule DBConnection.Holder do
...
+    :ets.update_element(sink_holder, :conn, [{conn(:state) + 1, state}])
+    :ok
+  end
+
+  @spec status?(pool_ref :: any, :ok | :aborted) :: boolean()
+  def status?(pool_ref(holder: holder), status) do
+    try do
+      :ets.lookup_element(holder, :conn, conn(:status) + 1) == status
+    rescue
+      ArgumentError -> false
```
fishcakez: does false make sense if `status` is aborted? See earlier comment so may not matter. Seems like an additional status.

lib/db_connection.ex
```diff
@@ -1445,73 +1378,67 @@ defmodule DBConnection do
...
   end
 
-  defp fail(conn) do
-    case get_info(conn, nil) do
-      {:ok, conn_state, _meter} ->
-        put_info(conn, :failed, conn_state)
-      _ ->
-        :ok
+  defp fail(%DBConnection{pool_ref: pool_ref}) do
+    case Holder.status?(pool_ref, :ok) do
```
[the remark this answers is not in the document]
josevalim: The reason why I don't want to return the status is because there are two external status that we would have to handle:

1. the done status
2. the missing ets status

So having this as ok? and aborted? is better IMO but the current code is also fine, especially since we have put_status elsewhere.

lib/db_connection.ex
josevalim: >  Does inlining keep the calling function in the stack even when bad_return! is called in the tail?

It does. This is because there may have been cases were `bad_return!` was called outside of the tail and this keeps it more future proof.

lib/db_connection/holder.ex
josevalim: I will refactor this in the separate PR where we improve the errors.

lib/db_connection/holder.ex
josevalim: In my mind, the missing ETS entry is equivalent to the pdict being removed. And all of the times we call `Holder.status?`we are not interested in those cases, hence false is reasonable.

lib/db_connection.ex
```diff
@@ -639,42 +637,47 @@ defmodule DBConnection do
...
+          case run(conn, &run_status/3, nil, opts) do
+            {^old_status, _meter} ->
+              {:ok, result}
+            {new_status, _meter} ->
+              {:nomatch, old_status, new_status}
+            {:error, _err, _meter} ->
+              {:error, result}
+            {kind, reason, stack, _meter} ->
+              :erlang.raise(kind, reason, stack)
+          end
```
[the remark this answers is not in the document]
josevalim: I will investigate this after merging.

josevalim:

lib/db_connection/holder.ex
```diff
@@ -46,30 +65,106 @@ defmodule DBConnection.Holder do
     end
   end
 
-  @spec checkin(pool_ref :: any, state :: any) :: :ok
-  def checkin(pool_ref, state) do
+  @spec checkin(pool_ref :: any) :: :ok
+  def checkin(pool_ref) do
     now = System.monotonic_time(@time_unit)
-    done(pool_ref, :checkin, state, now)
+    done(pool_ref, :ok, :checkin, now)
```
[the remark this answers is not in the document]
josevalim: Good catch. This can happen, I have verified by running the suite. I will open up a new issue.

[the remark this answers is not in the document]
josevalim: Possible fix in #163.

PR #156, opened 2018-09-22
Add pre_checkin and post_checkout to ownership

josevalim:
Note this is built on top of the jv-no-pdict branch.

### DIFF
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -46,20 +46,45 @@ defmodule DBConnection.Holder do
     end
   end
 
-  @spec checkin(pool_ref :: any, state :: any) :: :ok
-  def checkin(pool_ref, state) do
+  @spec checkin(pool_ref :: any) :: :ok
+  def checkin(pool_ref, _state \\ :TODO) do
     now = System.monotonic_time(@time_unit)
-    done(pool_ref, :checkin, state, now)
+    done(pool_ref, :checkin, now)
   end
 
-  @spec disconnect(pool_ref :: any, err :: Exception.t(), state :: any) :: :ok
-  def disconnect(pool_ref, err, state) do
-    done(pool_ref, :disconnect, state, err)
+  @spec disconnect(pool_ref :: any, err :: Exception.t()) :: :ok
+  def disconnect(pool_ref, err, _state \\ :TODO) do
+    done(pool_ref, :disconnect, err)
   end
 
-  @spec stop(pool_ref :: any, err :: Exception.t(), state :: any) :: :ok
-  def stop(pool_ref, err, state) do
-    done(pool_ref, :stop, state, err)
+  @spec stop(pool_ref :: any, err :: Exception.t()) :: :ok
+  def stop(pool_ref, err, _state \\ :TODO) do
+    done(pool_ref, :stop, err)
+  end
+
+  @spec handle(pool_ref :: any, fun :: atom, args :: [term], Keyword.t) :: tuple
+  def handle(pool_ref, fun, args, opts) do
+    pool_ref(holder: holder) = pool_ref
+
+    try do
+      :ets.lookup(holder, :conn)
+    rescue
+      e in ArgumentError ->
+        {:catch, :error, e, System.stacktrace()}
+    else
+      [conn(module: module, state: state)] ->
+        try do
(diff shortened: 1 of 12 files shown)


lib/db_connection/holder.ex
```diff
@@ -35,9 +35,9 @@ defmodule DBConnection.Holder do
 
   @spec delete(t) :: term
   def delete(holder) do
-    state = :ets.lookup_element(holder, :conn, conn(:state) + 1)
+    [conn(module: module, state: state)] = :ets.lookup(holder, :conn)
     :ets.delete(holder)
-    state
+    {module, state}
```
fishcakez: Spec of this fun should be update to match the return `{module, term}`

lib/db_connection/ownership/proxy.ex
```diff
@@ -157,32 +165,60 @@ defmodule DBConnection.Ownership.Proxy do
...
-      done.(pool_ref, err)
+      {conn_mod, conn_state} = Holder.delete(holder)
+
+      case pre_checkin.(op, conn_mod, conn_state) do
+        {:ok, ^original_mod, conn_state} ->
+          Holder.put_state(pool_ref, conn_state)
+          done.(pool_ref, err)
+          {:stop, {:shutdown, err}, state}
+
+         {:error, err, ^original_mod, conn_state} ->
```
fishcakez: Should this be `:disconnect` instead of `:error` as it disconnects?

lib/db_connection/ownership/proxy.ex
```diff
@@ -157,32 +165,60 @@ defmodule DBConnection.Ownership.Proxy do
...
   defp down(reason, state) do
     err = DBConnection.ConnectionError.exception(reason)
     pool_disconnect(err, state)
   end
 
+  defp log_and_disconnect(reason, state) do
+    err = DBConnection.ConnectionError.exception(reason)
+
+    _ = Logger.error(fn() ->
+      ["Ownership connection (", inspect(self()), ") disconnected: "
```
fishcakez: Will this be too noisy as the connection process will log to? Is this in case pre_checkin has an issue (crashes) and we want to see this first?

lib/db_connection/ownership/proxy.ex
```diff
@@ -30,40 +31,41 @@ defmodule DBConnection.Ownership.Proxy do
     timeout = Keyword.get(pool_opts, :queue_target, @queue_target) * 2
     interval = Keyword.get(pool_opts, :queue_interval, @queue_interval)
 
+    pre_checkin = Keyword.get(pool_opts, :pre_checkin, fn _, mod, state -> {:ok, mod, state} end)
+    post_checkout = Keyword.get(pool_opts, :post_checkout, &{:ok, &1, &2})
```
fishcakez: Do we definitely want to go with funs and not a behavior? As this is for testing only I think funs are preferred but if anyone would use this in prod we should switch to support code reload. I am very doubtful thats the case though. Either way we need to document these keyword values.

lib/db_connection/ownership/proxy.ex
josevalim: We were not logging anything before this change. I found this out while writing the new tests. So it is not related to pre_checkin but it was necessary (it was even causing failures in Ecto suite).

We log in the following cases:

1. owner exit while client is running
2. owner timeout
3. client timeout

I think it is important to log in those cases because they will lead to cascading changes and we may lose the root cause when those happen. For which case above are you worried about excessive messages? All three of them?

Note there are cases we don't log: client exit and owner exit without client.

lib/db_connection/ownership/proxy.ex
josevalim: Yeah, I am still sold on the funs because I find the behaviour too much overhead for something we don't know has a use case without testing. I will add the docs and leave a note.

josevalim:
Merged and added an issue to make sure we follow up in the relevant comments.

lib/db_connection/ownership/proxy.ex
josevalim: Actually, you are correct, we are logging twice in db connection tests. However, we don't get the logs in ecto tests. I am investigating.

josevalim: I will send a PR and continue the discussion there.

PR #159, opened 2018-09-24
Remove duplicate proxy logs

josevalim:
I was able to verify the logs appear twice in the DBConnection suite. However, once I remove those log calls, [this Ecto test no longer passes](https://github.com/elixir-ecto/ecto/blob/a17bf6089b539c45d8681054766b532f63e0a36d/integration_test/sql/sandbox.exs#L144) as nothing is logged. Thoughts @fishcakez?


josevalim:
I figured out why the error in ecto was no logged. It was because there was no proper connection checkout.

PR #166, opened 2018-09-27
Update docs

wojtekmach:


lib/db_connection.ex
```diff
@@ -804,19 +806,20 @@ defmodule DBConnection do
   ### Options
 
   See module documentation. The pool and connection module may support other
-  options. All options are passed to `handle_status/2`.
+  options. All options are passed to `c:handle_status/2`.
 
   ### Example
 
       {:ok, conn, _result} = DBConnection.begin(pool)
```
wojtekmach: there's neither `DBConnection.begin` nor `DBConnection.commit` below in the example, need to update this.

lib/db_connection.ex
```diff
@@ -134,7 +134,7 @@ defmodule DBConnection do
 
   This callback is called when the control of the state is passed back
   to the connection process. It should reverse any changes made in
-  `checkout/2`.
+  `c:checkout/1`.
```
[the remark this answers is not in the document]
josevalim: Yes.

lib/db_connection.ex
josevalim: It should be `DBConnection.transaction`.

josevalim:

lib/db_connection.ex
florinpatrascu: Hi there - without `begin`/`commit`, how would one use non-closure based transactions, with the new [DBConnection.html.transaction/3](
https://hexdocs.pm/db_connection/2.0.0-rc.0/DBConnection.html#transaction/3)? This is a follow up to #87, which is closed now. Thank you!

josevalim: It is not possible to have non-closure based transactions. It adds to many complications and possible failure scenarios, so we decided to not support it.

PR #171, opened 2018-10-01
Lock connection and pool

### LINKED ISSUE #169: handle should check the lock value
The lock is stored on pool_ref and matched after reading the entry on handle. If it doesn't, it can raise a runtime error.

josevalim:
Closes #169

### DIFF
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -6,8 +6,8 @@ defmodule DBConnection.Holder do
   @timeout 15000
   @time_unit 1000
 
-  Record.defrecord(:conn, [:connection, :module, :state, deadline: nil, status: :ok])
-  Record.defrecord(:pool_ref, [:pool, :reference, :deadline, :holder])
+  Record.defrecord(:conn, [:connection, :module, :state, :lock, deadline: nil, status: :ok])
+  Record.defrecord(:pool_ref, [:pool, :reference, :deadline, :holder, :lock])
 
   @type t :: :ets.tid()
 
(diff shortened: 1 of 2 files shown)


lib/db_connection/holder.ex
```diff
@@ -218,21 +221,25 @@ defmodule DBConnection.Holder do
...
 
     receive do
-      {:"ETS-TRANSFER", holder, pool, {^mref, ref}} ->
-        Process.demonitor(mref, [:flush])
-        deadline = start_deadline(timeout, pool, ref, holder, start)
-        pool_ref = pool_ref(pool: pool, reference: ref, deadline: deadline, holder: holder)
+      {:"ETS-TRANSFER", holder, pool, {^lock, ref}} ->
+        Process.demonitor(lock, [:flush])
+        {deadline, ops} = start_deadline(timeout, pool, ref, holder, start)
+        :ets.update_element(holder, :conn, [{conn(:lock) + 1, lock} | ops])
```
fishcakez: We need to release the lock in `done`, which can be done by adding an op to set it to nil.

I think can test that one by doing a disconnect, send message to caller in disconnect callback then blocking, trying to use the conn again, then send message to unblock disconnect.

josevalim: @fishcakez I came up with a simpler test. We can just try to do a execute with a stale connection we got from a run outside of another run. You can see it in this commit: 583bc5c35ce1af6d95e3853d698b015b5e41910b

You can see that I only remove the lock on checkin as we may try to do handle operations after disconnect, such as handle_status and handle_rollback, but those operations are not actually invoked as they hit the error/aborted branches of handle.

We could clean up the lock on disconnect/stop too and rearrange the handle clauses - but I found this approach cleaner. Let me know your thoughts.

josevalim:

PR #175, opened 2018-11-16
Cancel after_connect timeout once the connection is transferred

josevalim:


darrenclark:
I gave this fix a quick test in my project & it seems to have fixed the problem 👍

josevalim:

PR #179, opened 2018-11-29
queue config doc fix: milliseconds, not micro

### LINKED ISSUE #178: queue_target / queue_interval: milliseconds, or microseconds?
[The docs for queue config](https://hexdocs.pm/db_connection/DBConnection.html#start_link/2-queue-config) currently state (formatting added):

> Handling requests is done through a queue. When DBConnection is  started, there are two relevant options to control the queue:
> 
>   * `:queue_target` in **microseconds**, defaults to 50
>   * `:queue_interval` in **microseconds**, defaults to 1000
> 
>  Our goal is to stay under `:queue_target` for `:queue_interval`. In case we can't reach that, then we double the :queue_target. If we go above that, then we start dropping messages. For example, by default our queue time is 50**ms**. If we stay above 50**ms** for a whole secnod, we double the target to 100**ms** and we start dropping messages once it goes above the new limit.

The list specifically says **microseconds** with a default of 50, but down below it says **ms** which is the abbreviation for **milliseconds**. Which is correct? `µs` would be the correct abbreviation if microseconds was correct.

Also note the typo `secnod`. I would fix that on its own but figure it can be done at the same time as the more important doc fix this issue is about.

bgentry:
Fixes #178.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -362,15 +362,15 @@ defmodule DBConnection do
   Handling requests is done through a queue. When DBConnection is
   started, there are two relevant options to control the queue:
 
-    * `:queue_target` in microseconds, defaults to 50
-    * `:queue_interval` in microseconds, defaults to 1000
+    * `:queue_target` in milliseconds, defaults to 50
+    * `:queue_interval` in milliseconds, defaults to 1000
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #180, opened 2018-12-13
Update docs

wojtekmach:
- handle_declare no longer can return {:ok, cursor, state}
- handle_prepare no longer can return {:ok, result, state}
- handle_rollback can return any status
- Re-use status type in handle_status


wojtekmach:
There's at least one other callback that needs update, I'll re-open when done.

wojtekmach:
ready!

josevalim:

PR #188, opened 2019-02-19
Add start_option/0 and option/0 types

wojtekmach:
See https://github.com/elixir-ecto/myxql/pull/41/files for how it can be used. When this is merged, I'll perform similar changes in Postgrex.

Latest versions on Hex are db_connection 2.0.5 and postgrex 0.14.1. Say we release these types on 2.0.6 and 0.14.2, respectively. If users update to Postgrex 0.14.2 but keeps db_connection < 2.0.6, they will get a Dialyzer error:

    Unknown type: DBConnection.start_option/0

We could bump Postgrex's dependency requirement but that seems too much? Perhaps having this comment here should be enough to unblock folks if they run into it.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -103,6 +103,28 @@ defmodule DBConnection do
   @type cursor :: any
   @type status :: :idle | :transaction | :error
 
+  @type start_option ::
+          {:after_connect, (t -> any) | {module, atom, [any]} | nil}
+          | {:after_connect_timeout, timeout}
+          | {:backoff_max, non_neg_integer}
+          | {:backoff_min, non_neg_integer}
+          | {:backoff_type, :stop | :exp | :rand | :rand_exp}
+          | {:configure, (keyword -> keyword) | {module, atom, [any]} | nil}
+          | {:idle_interval, non_neg_integer}
+          | {:max_restarts, non_neg_integer}
+          | {:max_seconds, pos_integer}
+          | {:name, GenServer.name()}
+          | {:pool, module}
+          | {:pool_size, pos_integer}
+          | {:queue_interval, non_neg_integer}
+          | {:queue_target, non_neg_integer}
+          | {:show_sensitive_data_on_connection_error, boolean}
+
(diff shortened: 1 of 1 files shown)


josevalim:
Folks running dialyzer can jump to the latest, yeah.

lib/db_connection.ex
```diff
@@ -103,6 +103,28 @@ defmodule DBConnection do
...
+          | {:pool, module}
+          | {:pool_size, pos_integer}
+          | {:queue_interval, non_neg_integer}
+          | {:queue_target, non_neg_integer}
+          | {:show_sensitive_data_on_connection_error, boolean}
+
+  @type option ::
+          {:log, (DBConnection.LogEntry.t -> any) | {module, atom, [any]} | nil}
+          | {:queue, boolean}
+          | {:timeout, timeout}
```
fishcakez: also `{:deadline, integer}`

wojtekmach:
@fishcakez done. I've also took a stab at documenting this option in one of the functions, if this looks good I think we should repeat it in other functions as we do for `:timeout` - let me know.

lib/db_connection.ex
```diff
@@ -423,6 +446,9 @@ defmodule DBConnection do
     `start_link/2` docs
     * `:timeout` - The maximum time that the caller is allowed to perform
     this operation (default: `15_000`)
+    * `:deadline` - If set, specifies absolute monotonic time in milliseconds
+    by which caller must perform operation. See `System` module documentation
+    for more information on monotonic time
```
fishcakez: `default: nil` and overrides `timeout` if set

wojtekmach: ok how about this?

```
* `:deadline` - If set, overrides `:timeout` option and specifies absolute monotonic time in milliseconds
by which caller must perform operation. See `System` module documentation
for more information on monotonic time (default: `nil`)
```

I'll update the type to be `{:deadline, integer | nil}` too.

fishcakez: 👍

PR #189, opened 2019-02-19
Add :ping_timeout option

wojtekmach:
Ref: https://github.com/elixir-ecto/myxql/issues/38#issuecomment-459985140

This is WIP and it's missing docs & tests. It also fails the build as the integration test expects that ping callback is executed in the very same process as connect callback is.

@fishcakez some guidance would be appreciated! Should I look into timers and using DBConnection.Task.start_child as we're doing with after_connect?

### DIFF
--- a/lib/db_connection/connection.ex
+++ b/lib/db_connection/connection.ex
@@ -52,7 +52,8 @@ defmodule DBConnection.Connection do
       timer: nil,
       backoff: Backoff.new(opts),
       after_connect: Keyword.get(opts, :after_connect),
-      after_connect_timeout: Keyword.get(opts, :after_connect_timeout, @timeout)
+      after_connect_timeout: Keyword.get(opts, :after_connect_timeout, @timeout),
+      ping_timeout: Keyword.get(opts, :ping_timeout, 1_000)
     }
 
     {:connect, :init, s}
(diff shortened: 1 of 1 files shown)


lib/db_connection/connection.ex
```diff
@@ -52,7 +52,8 @@ defmodule DBConnection.Connection do
       timer: nil,
       backoff: Backoff.new(opts),
       after_connect: Keyword.get(opts, :after_connect),
-      after_connect_timeout: Keyword.get(opts, :after_connect_timeout, @timeout)
+      after_connect_timeout: Keyword.get(opts, :after_connect_timeout, @timeout),
+      ping_timeout: Keyword.get(opts, :ping_timeout, 1_000)
```
fishcakez: Should be `@timeout` default

lib/db_connection/connection.ex
```diff
@@ -381,4 +376,29 @@ defmodule DBConnection.Connection do
         [{mod, fun, length(args), info} | rest]
     end
   end
+
+  defp handle_ping(state, s) do
+    %{mod: mod, ping_timeout: timeout} = s
+
+    task =
+      Task.async(fn ->
```
fishcakez: I think we should use the task pool because it handles test cases and ensures semantics. When the task pool was introduced we needed a non-blocking interface but I think we can simplify things greatly by moving that pool to a similar async/yield model. Lets use a supervisor and `async_nolink` though. Perhaps refactor after connect to that model first, then look at using it for ping. Any functions should be inside a DBConnection.run call if outside the connection process, and we need to be using `Holder` too.

josevalim:
Closing this due to the complexity  and the fact implementation on the socket side is trivial.

PR #190, opened 2019-02-20
Track caller on proxy checkin

josevalim:


lib/db_connection/ownership/manager.ex
```diff
@@ -66,29 +66,29 @@ defmodule DBConnection.Ownership.Manager do
...
-    share_and_reply(state, pid)
+  def handle_call({:mode, {:shared, shared}}, {caller, _}, state) do
+    share_and_reply(state, shared, caller)
   end
   def handle_call({:mode, mode}, _from, %{mode: mode} = state) do
     {:reply, :ok, state}
   end
-  def handle_call({:mode, mode}, _from, state) do
-    state = proxy_checkin_all(state, nil)
+  def handle_call({:mode, mode}, {caller, _from}, state) do
```
fishcakez: nit: `_from` is not a good var name here as its only the `tag`, I think `_` more suitable

lib/db_connection/ownership/proxy.ex
```diff
@@ -36,7 +36,7 @@ defmodule DBConnection.Ownership.Proxy do
 
     state = %{client: nil, timer: nil, holder: nil,
               timeout: timeout, interval: interval, poll: nil,
-              owner_ref: owner_ref, pool: pool, pool_ref: nil,
+              owner: {caller, owner_ref}, pool: pool, pool_ref: nil,
```
fishcakez: This could be another entry in the map because this data never changes. `client` has a tuple to ensure "atomicity" so that we always set the whole client status and should be less likely to make a mistake. For owner/owner ref we never check both at same time so code would be clearer without the tuple.

josevalim: Yeah, but I also thought that if we were ever to change it, keeping them as a tuple guarantee we won't make stuff inconsistent. 

But I am indifferent. If you have a strong preference for one or another please let me know and I will change it.

lib/db_connection/ownership/manager.ex
josevalim:
```suggestion
  def handle_call({:mode, mode}, {caller, _}, state) do
```

josevalim:
I think we should also include which code the caller is running. It will make debugging much easier.

josevalim:
Just pushed caller stacktrace as well.

josevalim:

PR #196, opened 2019-05-21
Add process info stacktrace, if available, to logged message

### LINKED ISSUE #195: Improve context for timeout errors logged
We're seeing a randomly occurring error (does not correlate to slow queries or increased load on our database server):

```
Postgrex.Protocol (#PID<0.393.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.23831.49> timed out because it queued and checked out the connection for longer than 30000ms
```

Which I believe comes from `ConnectionPool`'s timeout handling. Unfortunately, there is no record of this in our slow query logs.

I'm not sure which direction to take when debugging further. If possible, could more information be added that might help direct us to the cause of this timeout? Perhaps some sort of backtrace or more information about the processes with these PIDs?

Is it possible that this log is a mistake? I see there is a check for the current holder:
> Check that timeout refers to current holder (and not previous)

tpitale:
Resolve #195

### DIFF
--- a/lib/db_connection/connection.ex
+++ b/lib/db_connection/connection.ex
@@ -215,8 +215,16 @@ defmodule DBConnection.Connection do
       "client #{inspect pid} timed out because it checked out " <>
         "the connection for longer than #{timeout}ms"
 
-    exception = DBConnection.ConnectionError.exception(message)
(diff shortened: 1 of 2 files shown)


josevalim:

tpitale:
Thank you @josevalim!

PR #201, opened 2019-07-10
Improve broken sentence

obrok:
I found this sentence that didn't quite make sense in the docs. I fixed according to what I think was meant by it, might still be wrong.


josevalim:

PR #203, opened 2019-07-23
Document testing a little bit

cdegroot:
WISOTT. All the non-standard stuff blindsided me so that probably means it should be documented :)


josevalim:
I have pushed something else meanwhile. Thanks for the PR.

PR #204, opened 2019-09-03
Logger with crash_reason metadata

tpitale:
I'm adding `crash_reason` on use of `Logger.error` so that projects using libraries like `sentry-elixir` will be able to capture these logged messages.


tpitale:
Ugh … auto-formatting. Let me fix.

lib/db_connection/ownership/proxy.ex
```diff
@@ -3,7 +3,6 @@ defmodule DBConnection.Ownership.Proxy do
 
   alias DBConnection.Holder
   use GenServer, restart: :temporary
-  require Logger
```
tpitale: I don't think this `Logger` is used.

josevalim:

PR #206, opened 2019-09-07
Ping all stale connections on idle interval

josevalim:
One possible downside of this approach is that
we may shut down all connections at once and if
there is a request around this time, the response
time will be higher. However, this is likely better
than the current approach, where we ping only the
first one, which means we can have a pool of stale
connections. Ideally we would introduce some sort
of jitter, to ensure we don't disconnect all of
them immediately, but this can be in the future.
The current behaviour is the same as in v1.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v2.2.0
+
+* Enhancements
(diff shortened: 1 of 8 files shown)


josevalim:

abashir17:
Hello @josevalim ,

Is there an ETA for when this version is going to be released?

PR #207, opened 2019-09-14
Improve docs around DBConnection.Query

wojtekmach:
cc @josevalim @fishcakez


wojtekmach:
So just to be clear, both DBConnection.encode and handle_execute are executed in the caller process, it's just that when we raise an error from the handle_execute callback we run into the issue described here: https://github.com/elixir-ecto/myxql/issues/76#issuecomment-531472364

josevalim:

PR #210, opened 2019-11-25
Add idle_time to LogEntry

### LINKED ISSUE #209: Insight into number of idle connections
We have telemetry for queue time which is very helpful, and we can increase the size of the connection pool based on this data. However, decreasing the amount of connections is still trial-and-error. It would be nice if there was a way to see how many connections are idle at a given time. We can then poll this every second or so, and be more confident about changing the pool size.

josevalim:
Closes #209.


josevalim:

PR #218, opened 2020-05-16
Add raising reason to ConnectionError metadata

qcam:


lib/db_connection/connection.ex
```diff
@@ -1,5 +1,17 @@
 defmodule DBConnection.ConnectionError do
-  defexception [:message, severity: :error]
+  defexception [:message, severity: :error, metadata: []]
```
josevalim, line 2: Let's go with:

```suggestion
  defexception [:message, severity: :error, reason: :error]
```

WDYT?

lib/db_connection/connection.ex
```diff
@@ -1,5 +1,17 @@
 defmodule DBConnection.ConnectionError do
-  defexception [:message, severity: :error]
+  defexception [:message, severity: :error, reason: :error]
+
+  @moduledoc """
+  The raised exception might include the reason which would be useful
```
qcam, line 5: I'm not sure if this is the best way to document exceptions. Please feel free to suggest any better way.

josevalim:

PR #219, opened 2020-06-17
DBConnection.Watcher should handle unknown messages

Cantido:
This is a good practice because anything can send our Watcher process a message and crash it.
From my own experimentation, the process isn't starting back up, either.

Now, I should say that I _do_ have a process that sends messages to random processes,
as part of a chaos engineering effort on my team, so it's up to y'all if this is a practical concern or not.
From what I understand, though, implementing a catch-all `handle_info` clause is a best practice.

### DIFF
--- a/lib/db_connection/watcher.ex
+++ b/lib/db_connection/watcher.ex
@@ -48,6 +48,10 @@ defmodule DBConnection.Watcher do
   def handle_info({:EXIT, _, _}, state) do
     {:noreply, state}
   end
+  
+  def handle_info(_, state) do
+    {:noreply, state}
+  end
 
   def terminate(_, {_, started_refs}) do
     for {_, {caller_pid, _}} <- started_refs do


josevalim:
Hi @Cantido, thanks for the PR!

>  From what I understand, though, implementing a catch-all handle_info clause is a best practice.

Not necessarily. For example, I often prefer to not implement a catch-all because if it does receive a message that it doesn't expect, then it may be a bug somewhere in the software (i.e. I forgot to receive a message that I should OR someone is sending a message to the wrong place).

For the reason above, I prefer to not merge this one for now. Thanks!

Cantido:
Understood! Thanks for the quick response!

Cantido:
I have a related question, actually. When this process crashes, I'm not seeing it restart. In fact, it seems to take down the entire DB connection and it doesn't reconnect. Queries fail with `** (DBConnection.ConnectionError) tcp recv: closed`. Is this intended?

josevalim:
I need to investigate to be sure. :)

Cantido:
If you could do that, that would be fantastic! Either way, I got some good results from this experiment of mine and my app will be much more tolerant of DB errors.

I super appreciate the work you do. Thank you!

PR #224, opened 2020-09-26
Replace deprecated stacktraces and bump requirements

sabiwara:
Replace deprecated `System.stacktrace()` by `__STACKTRACE__` to fix warnings in Elixir 1.11.
Bumped minimal elixir version accordingly to 1.7 (like ecto's [current release](https://github.com/elixir-ecto/ecto/blob/v3.4.6/mix.exs#L10))

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -717,9 +717,8 @@ defmodule DBConnection do
           {result, run(conn, &run_status/3, nil, opts)}
         catch
           kind, error ->
-            stacktrace = __STACKTRACE__
             checkin(conn)
-            :erlang.raise(kind, error, stacktrace)
+            :erlang.raise(kind, error, __STACKTRACE__)
         else
           {result, {:error, _, _}} ->
             checkin(conn)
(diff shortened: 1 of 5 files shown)


lib/db_connection.ex
```diff
@@ -717,7 +717,7 @@ defmodule DBConnection do
           {result, run(conn, &run_status/3, nil, opts)}
         catch
           kind, error ->
-            stacktrace = System.stacktrace()
+            stacktrace = __STACKTRACE__
```
michalmuskala, line 720: With the `__STACKTRACE__` construct we don't need to capture it in a variable anymore, we can inline it - there's no risk of capturing the wrong stacktrace like with `System.stacktrace/0`

sabiwara: Thanks, will fix!

sabiwara:
Side note for the record: there are also warnings from the underlying `connection` library that could be fixed by bumping once this fix is released: https://github.com/fishcakez/connection/pull/13

josevalim:

Issue #227, opened 2020-11-17
Add support for an `after_disconnect` callback function


slashdotdash:
It can sometimes be useful to know when a database connection has disconnected. For instance, using [Postgres advisory locks](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS) which are tied to a session means that the lock is released on disconnect. Another example might be to change application behaviour when a database connection is unavailable. 

Therefore it would be useful to support an `after_disconnect` callback function passed in the `DBConnection.start_link/2` function options which behaves similarly to the existing `after_connect` option. The existing function is called after connecting, I propose we add a corresponding function after disconnecting. 

Does this sound like a reasonable feature to add before I attempt to work on a pull request?

josevalim:
Hi @slashdotdash!

This is tricky because what if the connection was disconnected because the connection process crashed? So I think we need a mixture of event listeners (a process db_connection will send a message to whenenever it connects/disconnects) and you also have to setup monitors on our side. A PR for the first part will be welcome.

slashdotdash:
@josevalim Would it be sufficient to pass the `after_disconnect` function to the `DBConnection.Connection` process to have it handle disconnects and `DBConnection.Watcher` process to handle connection exits?

josevalim:
I would prefer to not allow custom code to run inside the connection process. Even for `after_connect`, we start a separate process and it is somewhat complicated. Is there a reason why a listener approach won't do?

slashdotdash:
Can we discuss this further as I'm unsure how the listener approach might work. What's the best way to talk it through?

josevalim:
Either here or on #elixir-lang channel on freenode.

PR #228, opened 2020-11-17
Add `:connection_listeners` option to `DBConnection.start_link/2`

slashdotdash:
The `:connection_listeners` option allows one or more processes to be notified whenever a connection is connected or disconnected. A listener may be a remote or local PID, a local port, a locally registered name, or a tuple in the form of `{registered_name, node}` for a registered name at another node.

Each listener process may receive the following messages where `pid` identifies the connection process:

* `{:connected, pid}`
* `{:disconnected, pid}`

Related to #227.

@josevalim Does this follow what you were initally suggesting?

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -105,6 +105,7 @@ defmodule DBConnection do
   @type start_option ::
           {:after_connect, (t -> any) | {module, atom, [any]} | nil}
           | {:after_connect_timeout, timeout}
+          | {:connection_listeners, list(Process.dest()) | nil}
           | {:backoff_max, non_neg_integer}
           | {:backoff_min, non_neg_integer}
           | {:backoff_type, :stop | :exp | :rand | :rand_exp}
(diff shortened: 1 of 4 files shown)


lib/db_connection.ex
```diff
@@ -410,6 +414,20 @@ defmodule DBConnection do
   requests before they are sent to the database, which would
   otherwise increase the burden on the database, making the
   overload worse.
+
+  ## Connection listeners
+
+  The `:connection_listeners` option allows one or more processes to be notified
+  whenever a connection is connected or disconnected. A listener may be a remote
+  or local PID, a local port, a locally registered name, or a tuple in the form
```
josevalim, line 422: It can't really be a port as we don't send messages required by the port format. :)

josevalim:
@slashdotdash just one minor comment, everything else looks excellent. The biggest question now is if this is enough to solve your original problems. :)

lib/db_connection.ex
slashdotdash: I've removed port from the list in 0eae12ea98a49120a3634a3a9afc1e9a186701e9.

slashdotdash:
> The biggest question now is if this is enough to solve your original problems. :)

Yes, I think so. My goal was to allow a connection to be monitored so when it is connected or disconnected I can handle these state changes elsewhere. It was possible to use the `after_connect` callback to support the connected change, but not disconnects. The connection listeners allows both.

josevalim:

slashdotdash:
@josevalim I did some testing of the new `:connection_listeners` option and it worked as I had hoped for when simulating a flaky database connection with Postgrex by starting/stopping the Postgres server.

> {:connected, #PID<0.297.0>}
> {:disconnected, #PID<0.297.0>}
> {:connected, #PID<0.298.0>}

PR #229, opened 2020-11-18
Fix failing connection listeners test

slashdotdash:
Add a longer back-off minimum value to the connection listeners tests to allow for slower test execution.

Fixes failing CI tests in #228 when run for older Elixir and Erlang versions.


josevalim:

PR #232, opened 2020-12-25
Mix format

kianmeng:
Since the minimum Elixir version is 1.7+, we should mix format the
the whole module.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,4 @@
+# Used by "mix format"
+[
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+]
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -1,10 +1,7 @@
 defmodule DBConnection.Stream do
   defstruct [:conn, :query, :params, :opts]
 
-  @type t :: %__MODULE__{conn: DBConnection.conn,
-                         query: any,
-                         params: any,
-                         opts: Keyword.t}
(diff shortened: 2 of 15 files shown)


josevalim:

josevalim:
Thanks! I generally don't accept mix formatting PRs but all of the code formatted code here looked good without intervention, so :+1:.

kianmeng:
> Thanks! I generally don't accept mix formatting PRs but all of the code formatted code here looked good without intervention, so .

Noted. One question, for any legacy Elixir modules (created before Elixir 1.6), should we `mix format` it to update it to the latest code format or let it be?

wojtekmach:
I believe running mix format on existing projects is not really about whether they're using Elixir v1.7+ but more about code churn. For example, we likely wouldn't accept a PR to Ecto that mix formats everything because:

a) there's a ton of code and we'd likely require a lot manual interventions
b) it would mess up git history

kianmeng:
> I believe running mix format on existing projects is not really about whether they're using Elixir v1.7+ but more about code churn. For example, we likely wouldn't accept a PR to Ecto that mix formats everything because:
> 
> a) there's a ton of code and we'd likely require a lot manual interventions
> b) it would mess up git history

Noted and thanks for the explanation. If I encounter these type of legacy modules without code format, I will opt to raise a ticket instead and up to the module owner to mix format the whole repository or not.

michalmuskala:
Git history can be dealt with using the relatively new `--ignore-rev` option to `git blame`. 

There's a third reason, though, important for projects maintaining multiple versions, that back-porting fixes gets harder.

Issue #234, opened 2021-03-08
Telemetry event for capturing DBConnection.ConnectionError


polvalente:
Is there any way to measure through telemetry if the database keeps dropping connections with DBConnection.ConnectionError?

I noticed that the `drop` private function in `deps/db_connection/lib/db_connection/connection_pool.ex` does print a custom inspect whenever the connection is dropped due to timeout. Maybe that could be a place to add the event? Or is there a better place in :ecto or :ecto_sql to do so?

I could work on the PR myself given the right direction!

Thanks in advance

josevalim:
We can definitely add a telemetry event for that. And likely one for when connections are established too.

polvalente:
@josevalim I was reading #228 and maybe adding another message for timeouts could also be a way to handle this. What do you think?

josevalim:
That also sounds good to me! PRs are welcome!

polvalente:
> That also sounds good to me! PRs are welcome!

Ok then! I'll try to come up with _something_

polvalente:
Resolved in #236

PR #235, opened 2021-03-09
feat: send checkout timeout event to connection listeners

polvalente:
Adds the `:checkout_timeout` event as suggested in #234

My main concern here is that I had to use the ConnectionPool as a message relay for the `connection_listeners`.
Is there a way to avoid this? I'm worried that this could cause an inbox overflow to the ConnectionPool.

Another workaround would be to publish this as a Telemetry event, altough in that case I wouldn't know how to fetch the "telemetry event prefix" that is passed to Ecto.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -423,6 +423,7 @@ defmodule DBConnection do
 
     * `{:connected, pid}`
     * `{:disconnected, pid}`
+    * `{:queue_timeout, %DBConnection.ConnectionError{}}`
 
   """
   @spec start_link(module, opts :: Keyword.t()) :: GenServer.on_start()
--- a/lib/db_connection/connection_pool.ex
+++ b/lib/db_connection/connection_pool.ex
@@ -34,7 +34,8 @@ defmodule DBConnection.ConnectionPool do
       next: now_in_ms,
       poll: nil,
       idle_interval: idle_interval,
-      idle: nil
+      idle: nil,
+      connection_listeners: Keyword.get(opts, :connection_listeners) || []
     }
 
     codel = start_idle(now_in_native, start_poll(now_in_ms, now_in_ms, codel))
@@ -147,6 +148,14 @@ defmodule DBConnection.ConnectionPool do
   def handle_info({:timeout, idle, past_in_native}, {_, _, %{idle: idle}} = data) do
     {status, queue, codel} = data
     drop_idle(past_in_native, status, queue, codel)
+  end
+
+  def handle_info(
+        {:checkout_timeout, err},
+        {_, _, %{connection_listeners: connection_listeners}} = state
+      ) do
+    Enum.map(connection_listeners, &send(&1, {:checkout_timeout, self(), err}))
+    {:noreply, state}
   end
 
   defp drop_idle(past_in_native, status, queue, codel) do
--- a/lib/db_connection/connection_pool.ex
+++ b/lib/db_connection/connection_pool.ex
@@ -34,8 +34,7 @@ defmodule DBConnection.ConnectionPool do
       next: now_in_ms,
       poll: nil,
       idle_interval: idle_interval,
-      idle: nil,
-      connection_listeners: Keyword.get(opts, :connection_listeners) || []
+      idle: nil
     }
 
     codel = start_idle(now_in_native, start_poll(now_in_ms, now_in_ms, codel))
--- a/lib/db_connection/connection_pool.ex
+++ b/lib/db_connection/connection_pool.ex
@@ -150,14 +150,6 @@ defmodule DBConnection.ConnectionPool do
     drop_idle(past_in_native, status, queue, codel)
   end
 
-  def handle_info(
-        {:checkout_timeout, err},
-        {_, _, %{connection_listeners: connection_listeners}} = state
-      ) do
-    Enum.map(connection_listeners, &send(&1, {:checkout_timeout, self(), err}))
-    {:noreply, state}
-  end
-
   defp drop_idle(past_in_native, status, queue, codel) do
     # If no queue progress since last idle check oldest connection
     case :ets.first(queue) do
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -88,7 +88,7 @@ defmodule DBConnection.Holder do
       end
 
     case result do
-      {:error, %DBConnection.ConnectionError{} = err} ->
+      {:error, %DBConnection.ConnectionError{reason: :queue_timeout} = err} ->
         Enum.each(opts[:connection_listeners] || [], &send(&1, {:checkout_timeout, err}))
 
       _ ->
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -89,7 +89,7 @@ defmodule DBConnection.Holder do
 
     case result do
       {:error, %DBConnection.ConnectionError{} = err} ->
-        send(pool, {:checkout_timeout, err})
+        Enum.each(opts[:connection_listeners] || [], &send(&1, {:checkout_timeout, err}))
 
       _ ->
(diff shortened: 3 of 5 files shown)


lib/db_connection/connection_pool.ex
```diff
@@ -149,6 +150,14 @@ defmodule DBConnection.ConnectionPool do
     drop_idle(past_in_native, status, queue, codel)
   end
 
+  def handle_info(
```
polvalente, line 153: This is the part that I'm concert about. Too many `:checkout_timeout`s could lead to the ConnectionPool being overflowed with messages, unable to handle anything else.

Any thoughts on this?

josevalim:
You are correct. Perhaps it is best to use telemetry for this rather than the connection listener.

polvalente:
> You are correct. Perhaps it is best to use telemetry for this rather than the connection listener.

Any ideas on how to deal with the configurable prefix? I tried to look, but there didn't seem to be an easy way to find it.
Anyway, I'll start refactoring this to emit an event instead `send`ing the message

josevalim:
The configurable prefix has been generally discouraged in telemetry and mostly no other project uses it, besides Ecto for legacy reasons. My suggestion is to use a set prefix and pass any relevant information as metadata. I assume we will at least have access to the repository name from here? 🤔

polvalente:
@josevalim It seems that either I overlooked something, or some of my changes changed something. I `inspect`ed the `opts` to see if the repo was available, but to my surprise, the connection_listeners were there!

lib/db_connection/connection_pool.ex
```diff
@@ -34,7 +34,8 @@ defmodule DBConnection.ConnectionPool do
       next: now_in_ms,
       poll: nil,
       idle_interval: idle_interval,
-      idle: nil
+      idle: nil,
+      connection_listeners: Keyword.get(opts, :connection_listeners) || []
```
josevalim, line 38: I assume we don't need this change anymore?

polvalente: Yup. I've just reverted this :)

lib/db_connection/holder.ex
```diff
@@ -56,35 +56,46 @@ defmodule DBConnection.Holder do
...
+
+        {:exit, reason} ->
+          exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+      end
 
-      {:exit, reason} ->
-        exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+    case result do
+      {:error, %DBConnection.ConnectionError{} = err} ->
+        Enum.each(opts[:connection_listeners] || [], &send(&1, {:checkout_timeout, err}))
```
josevalim, line 92: Those errors are going to be other things besides checkout_timeout, no?

lib/db_connection/connection_pool.ex
polvalente: It seems that I really let this fly by me :sweat_smile:

lib/db_connection/holder.ex
josevalim: I think if you want only the checkout timeouts, then we need to put it close to where the error is actually raised?

polvalente: Yes. Maybe we can relay them all but change the message to a more relevant name?
Checkout timeouts have `reason: :queue_timeout` in the struct

polvalente: > I think if you want only the checkout timeouts, then we need to put it close to where the error is actually raised?

I'm not sure which is better. I can see use for having access to all the errors, but I'm almost sure that the other errors can be inferred through other Ecto Telemetry events... What do you think?

Also, I'd need some help in putting this closer to the error. Another option is having a more strict match in this `case` for identifying the relevant `:reason`

polvalente: Thinking as a maintainer, maybe relaying only `:queue_timeout`s would be better in the sense that adding features/events is easy, but removing them could be problematic

lib/db_connection/holder.ex
```diff
@@ -56,35 +56,46 @@ defmodule DBConnection.Holder do
...
+          end
+
+        {:exit, reason} ->
+          exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+      end
 
-      {:exit, reason} ->
-        exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+    case result do
+      {:error, %DBConnection.ConnectionError{} = err} ->
```
polvalente, line 91:
```suggestion
      {:error, %DBConnection.ConnectionError{reason: :queue_timeout} = err} ->
```

This would suffice for identifying timeouts

lib/db_connection/holder.ex
josevalim: > Thinking as a maintainer, maybe relaying only :queue_timeouts would be better in the sense that adding features/events is easy, but removing them could be problematic

Exactly. Although I am still wondering, why not emit this from the connection pool, which is where the error is raised?

polvalente: I couldn't my way well enough through the project's structure. In the exact place where the error is raised the connection_listeners info isn't available (lib/db_connection/connection_pool.ex:293 and lib/db_connection/ownership/proxy.ex:291), so I'd need to be a little higher in the callstack.

Could you point me towards the right direction?

polvalente: btw, I've already pushed the version which only relays `:queue_timeout`

lib/db_connection/holder.ex
```diff
@@ -56,35 +56,46 @@ defmodule DBConnection.Holder do
...
+
+        {:exit, reason} ->
+          exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+      end
 
-      {:exit, reason} ->
-        exit({reason, {__MODULE__, :checkout, [pool, opts]}})
+    case result do
+      {:error, %DBConnection.ConnectionError{reason: :queue_timeout} = err} ->
+        Enum.each(opts[:connection_listeners] || [], &send(&1, {:checkout_timeout, err}))
```
josevalim, line 92: Should we name the event queue_timeout for consistency?

josevalim:
Ok, so let's ship as is and just rename the event. We also need docs and we should be good to go!

lib/db_connection/holder.ex
polvalente: Sounds good :)

lib/db_connection.ex
```diff
@@ -423,6 +423,7 @@ defmodule DBConnection do
 
     * `{:connected, pid}`
     * `{:disconnected, pid}`
+    * `{:queue_timeout, %DBConnection.ConnectionError{}}`
```
josevalim, line 426: Oh, one last question, does the error include all relevant information, such as the options and repository that failed?

polvalente: No, but the options passed to the `checkout` function is something that we can easily add. The repo itself I'm not sure, but in the worst case scenario, multi-repo systems can have one listener for each repo.

Inspection `opts` prints:

```elixir
[
  agent: #PID<0.292.0>,
  parent: #PID<0.291.0>,
  backoff_start: 30000,
  queue_timeout: 10,
  queue_target: 10,
  queue_interval: 10,
  connection_listeners: [#PID<0.291.0>],
  pool: DBConnection.ConnectionPool,
  pool_size: 1
]
```

polvalente: No worries! I could see both ideas making sense, at least we explored both options :D
Should I refactor this to publish the telemetry event then?
Which metadata should I include?

josevalim: I would include the whole exception and the options too.

josevalim: you can send the telemetry one as a separate PR, so we have both as reference!

polvalente:
Closed in favor of #236

PR #236, opened 2021-03-10
feat: add telemetry event for queue timeouts

polvalente:
As discussed in #235 and #234, this PR adds a telemetry event for `DBConnection.ConnectionError`s.

There are some discussion points still:
1. Which `opts` should we add to `metadata`?
2. Is there any meaningful measurement we could add instead of `%{error: 1}`?
3. Although the tests pass as expected, when I used this code in a local project (pointing to the local path for `db_connection`), the `:connection_listeners` option is empty. Is this expected? This was the only way I could think of for monitoring a project with different `Ecto.Repo`s. Or maybe this event should be published by either `postgrex`, or `ecto_sql` or even `ecto`?

I chose to keep publishing the event basically in the same place as in #235 because the `opts` could be useful in handling multiple-Ecto.Repo projects.

### DIFF
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -60,6 +60,19 @@ defmodule DBConnection.Holder do
       {:ok, _, _, _, _} = ok ->
         ok
 
+      {:error, %DBConnection.ConnectionError{} = connection_error} = error ->
+        :telemetry.execute(
+          [:db_connection, :connection_error],
+          %{count: 1},
+          %{
+            error: connection_error,
+            pool: opts[:pool],
+            connection_listeners: opts[:connection_listeners]
+          }
+        )
+
+        error
+
       {:error, _} = error ->
         error
 
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -66,8 +66,7 @@ defmodule DBConnection.Holder do
           %{count: 1},
           %{
             error: connection_error,
-            pool: opts[:pool],
-            connection_listeners: opts[:connection_listeners]
+            opts: opts
           }
         )
 
--- a/mix.exs
+++ b/mix.exs
@@ -24,7 +24,7 @@ defmodule DBConnection.Mixfile do
 
   def application do
     [
-      applications: [:logger, :connection],
+      applications: [:logger, :connection, :telemetry],
       mod: {DBConnection.App, []}
     ]
   end
@@ -32,7 +32,8 @@ defmodule DBConnection.Mixfile do
   defp deps do
     [
       {:connection, "~> 1.0"},
-      {:ex_doc, ">= 0.0.0", only: :dev, runtime: false}
+      {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
+      {:telemetry, "~> 0.4"}
     ]
   end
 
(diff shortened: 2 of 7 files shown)


mix.exs
```diff
@@ -24,15 +24,16 @@ defmodule DBConnection.Mixfile do
...
       mod: {DBConnection.App, []}
     ]
   end
 
   defp deps do
     [
       {:connection, "~> 1.0"},
-      {:ex_doc, ">= 0.0.0", only: :dev, runtime: false}
+      {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
+      {:telemetry, "~> 0.4"}
```
josevalim, line 36:
```suggestion
      {:telemetry, "~> 0.4 or ~> 1.0"}
```

mix.exs
```diff
@@ -24,15 +24,16 @@ defmodule DBConnection.Mixfile do
 
   def application do
     [
-      applications: [:logger, :connection],
+      applications: [:logger, :connection, :telemetry],
```
josevalim, line 27:
```suggestion
      extra_applications: [:logger],
```

lib/db_connection/holder.ex
```diff
@@ -60,6 +60,19 @@ defmodule DBConnection.Holder do
...
 
+      {:error, %DBConnection.ConnectionError{} = connection_error} = error ->
+        :telemetry.execute(
+          [:db_connection, :connection_error],
+          %{count: 1},
+          %{
+            error: connection_error,
+            pool: opts[:pool],
+            connection_listeners: opts[:connection_listeners]
+          }
```
josevalim, line 71: I think you can just include all options. :)

```suggestion
          %{
            error: connection_error,
            opts: opts
          }
```

josevalim:
Looks great, I have added some comments and we can ship it!

josevalim:

v0idpwn:
I'd like to emit telemetry events from disconnects coming from client-side, too, like when the process that checked out a connection exits, causing a disconnect. Currently we are measuring this through logs, but it's not possible to identify which Repo had the issue. Would a PR be welcome? cc @josevalim @greg-rychlewski

josevalim:
Would you send a PR for both projects so we can compare them? I think this particular event we have in this PR cannot be emitted by Postgrex, can it?

v0idpwn:
Actually, I think the same applies here, the telemetry event would need to be emitted from db_connection.

I wrote adding https://github.com/elixir-ecto/db_connection/commit/4256384a22f5597a6d3f0dceb34212b51ad88d43 as a PoC (names could be discussed, etc). What strikes me is that it makes `:connection_listeners` look a bit redundant, since it's a more powerful alternative (at least for disconnects, and adding an event for connects would be trivial).

josevalim:
Couldn’t telemetry be emitted from the listeners then? Maybe we ship a listener that does telemetry?

v0idpwn:
This would forbid me to add metadata, as listeners only receive `:connected`/`:disconnected` + a tag.
I was thinking of the reverse path: making a telemetry handler that sends to the listeners. 

Note: I figured its possible to use the connection listeners for my use-case due to the _newish_ tags feature, so it might be superfluous to implement the telemetry event. But using telemetry would probably provide a better API 👀

PR #240, opened 2021-04-16
Add DBConnection.disconnect_all/3

### LINKED ISSUE #99: Add expiration timeout that limits max life-time of a connection
This timeout should be randomized to prevent sync disconnects.

josevalim:
Closes #99.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -453,6 +453,20 @@ defmodule DBConnection do
   end
 
   @doc """
+  Forces all connections in the pool to disconnect within the given interval.
+
+  Once this function is called, the pool will disconnect all of its connections
+  as they are checked in or as they are pinged. Checked in connections will be
+  randomly checked in within the given time interval. Pinged connections are
+  immediately disconnected - as they are idle (according to `:idle_interval`).
+  """
+  def disconnect_all(conn, interval, opts \\ []) when interval >= 0 do
+    pool = Keyword.get(opts, :pool, DBConnection.ConnectionPool)
+    interval = System.convert_time_unit(interval, :millisecond, :native)
+    pool.disconnect_all(conn, interval, opts)
+  end
+
+  @doc """
   Prepare a query with a database connection for later execution.
 
   It returns `{:ok, query}` on success or `{:error, exception}` if there was
--- a/integration_test/connection_pool/disconnect_all_test.exs
+++ b/integration_test/connection_pool/disconnect_all_test.exs
@@ -1,0 +1,83 @@
+defmodule TestPoolDisconnectAll do
+  use ExUnit.Case, async: true
+
+  alias TestPool, as: P
+  alias TestAgent, as: A
+  alias TestQuery, as: Q
+  alias TestResult, as: R
+
+  test "disconnect on checkin" do
+    stack = [
+      {:ok, :state},
+      {:ok, %Q{}, %R{}, :new_state1},
+      {:ok, %Q{}, %R{}, :new_state2},
+      {:ok, :dead_state},
+      {:ok, :final_state},
+      {:ok, %Q{}, %R{}, :final_state1},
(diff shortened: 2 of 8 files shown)


lib/db_connection.ex
```diff
@@ -452,6 +456,28 @@ defmodule DBConnection do
     pool.child_spec({conn_mod, opts})
   end
 
+  @doc """
+  Forces all connections in the pool to disconnect within the given interval.
+
+  Once this function is called, the pool will disconnect all of its connections
+  as they are checked in or as they are pinged. Checked in connections will be
+  randomly checked in within the given time interval. Pinged connections are
```
josevalim, line 464:
```suggestion
  randomly disconnected within the given time interval. Pinged connections are
```

waltfy:
@josevalim amazing, thank you for this!

lib/db_connection.ex
```diff
@@ -452,6 +456,28 @@ defmodule DBConnection do
...
+
+  If the connection has a backoff configured (which is the case by default),
+  disconnecting means an attempt at a new connection will be done immediately
+  after, without starting a new process for each connection. However, if backoff
+  has been disabled, the connection process will terminate. In such cases,
+  disconnecting all connections may cause the pool supervisor to restart
+  depending on the max_restarts/max_seconds configuration of the pool,
+  so you will want to set those carefully.
+  """
+  def disconnect_all(conn, interval, opts \\ []) when interval >= 0 do
```
visciang, line 475: Do we need the `opts` parameter? It seems unused down the call stack and in the tests.
Or it's for consistency with the signature of the other public functions in the module?

lib/db_connection.ex
```diff
@@ -452,6 +456,28 @@ defmodule DBConnection do
...
+  immediately disconnected - as they are idle (according to `:idle_interval`).
+
+  If the connection has a backoff configured (which is the case by default),
+  disconnecting means an attempt at a new connection will be done immediately
+  after, without starting a new process for each connection. However, if backoff
+  has been disabled, the connection process will terminate. In such cases,
+  disconnecting all connections may cause the pool supervisor to restart
+  depending on the max_restarts/max_seconds configuration of the pool,
+  so you will want to set those carefully.
+  """
```
visciang, line 474: Missing typespec:
`@spec disconnect_all(conn, non_neg_integer, opts :: Keyword.t()) :: :ok`

visciang (commented):
I've manually done some integration tests against postgres.
It works as expected.

What I've seen, is a strange thing on disconnect.
On disconnect we execute `Postgrex.disconnect/2`:

```elixir
  def disconnect(_, s) do
    sock_close(s)
    _ = recv_buffer(s)
    delete_parameters(s)
    queries_delete(s)
    cancel_request(s)    # <<<<<< !
    :ok
  end
```

The `cancel_request(s)` at the very end, after the initial `socket_close(s)`, issue a cancel request command to a non existing process postgres side:

```
2021-08-19 19:25:46.650 UTC [18382] LOG:  PID 18370 in cancel request did not match any process
```

Shouldn't it be before `socker_close()`? I've done a quick test ans it works as expected.

While this problem is not caused by this PR but only stressed by the new feature, it will be very annoying to see a lot of the above error logs postgres side if we periodically recycle all the connection in the pool.

lib/db_connection.ex
josevalim: Yes, just in case it is needed in the future.

lib/db_connection.ex
josevalim, line 474:
```suggestion
  """
   @spec disconnect_all(conn, non_neg_integer, opts :: Keyword.t()) :: :ok
```

visciang:
Great @josevalim, I've see the cancel_request() fix postgrex side has been merged.

Thank you!

josevalim:

PR #250, opened 2021-09-14
Improve grammar in error message

vereis:
I ran into this error while running tests on a different project and noticed the awkward `albeit it` which I don't think is correct.

I changed it to `although this` which is a bit simpler, but could otherwise be `albeit increasing resource consumption`.

WDYT?


josevalim:

PR #254, opened 2022-02-21
Document `DBConnection.ConnectionPool`

wojtekmach:
This way we can reference it in the docs without an ExDoc warning. (Postgrex already did that which triggered this change.) Just adding a very basic `@moduledoc`, all its functions are still hidden.


josevalim:

PR #255, opened 2022-03-03
Enable detecting connection type of a pool process

jonatanklosko:
This allows for determining if the given process is a connection pool and also figure out the underlying driver.

### DIFF
--- a/lib/db_connection/connection_pool.ex
+++ b/lib/db_connection/connection_pool.ex
@@ -12,6 +12,7 @@ defmodule DBConnection.ConnectionPool do
   @queue_interval 1000
   @idle_interval 1000
   @time_unit 1000
+  @connection_module_key :connection_module
 
   @doc false
   def start_link({mod, opts}) do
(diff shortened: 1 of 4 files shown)


lib/db_connection/connection_pool.ex
```diff
@@ -33,10 +34,38 @@ defmodule DBConnection.ConnectionPool do
...
+         {:dictionary, dictionary} <- Process.info(pid, :dictionary),
+         {:ok, module} <- fetch_from_dictionary(dictionary, @connection_module_key),
+         do: {:ok, module},
+         else: (_ -> :error)
+  end
+
+  defp to_pid(nil), do: :error
+  defp to_pid(pid) when is_pid(pid), do: {:ok, pid}
+  defp to_pid(name) when is_atom(name), do: name |> Process.whereis() |> to_pid()
+  defp to_pid(_other), do: :error
```
josevalim, line 54: I think you can use GenServer.whereis.

lib/db_connection/connection_pool.ex
```diff
@@ -33,10 +34,38 @@ defmodule DBConnection.ConnectionPool do
...
+
+  defp fetch_from_dictionary(dictionary, key) do
+    Enum.find_value(dictionary, :error, fn
```
josevalim, line 57: You can use the keyword module? Otherwise, consider using `List.keyfind`!

jonatanklosko: pdict is not a keyword list, as the keys may be `{:any, :thing}`. The advantage over `List.keyfind` is that if there's a `nil` value we also want it (in this case it doesn't matter though) :)

jonatanklosko:
@josevalim updated!

lib/db_connection.ex
```diff
@@ -1085,6 +1087,32 @@ defmodule DBConnection do
...
+  def register_as_pool(conn_module) do
+    Process.put(@connection_module_key, conn_module)
+  end
+
+  @doc """
+  Returns connection module used by the given connection pool process.
+
+  If the given process is not a connection pool, `:error` is returned.
+  """
+  @spec connection_module(pid() | atom()) :: {:ok, module()} | :error
```
jonatanklosko, line 1100: This should probably be `conn()`, but I'm not sure if we can handle `%DBConnection{}` reasonably.

PR #257, opened 2022-03-03
Change disconnect_all log severity to debug

btkostner:
Because `disconnect_all` is being explicitly called, the logging message doesn't need to be as verbose as it is. This changes the severity to debug. From @josevalim's [comment here](https://github.com/elixir-ecto/db_connection/issues/99#issuecomment-1058472198).


josevalim:
It seems to have broken tests, can you please take a look? :)

btkostner:
Whoops. I missed that when editing on GitHub. Should be good now :+1:

josevalim:

Issue #260, opened 2022-05-19
Turn `ConnectionError` into an error tuple (vs exception)


acco:
Hey team,

`DbConnection.ConnectionError`s happen. We obviously have worked to minimize, but there are some high-volume tables in our app that will run into this issue from time to time.

The problem is that unlike other errors, a `ConnectionError` is an exception. This complicates our control flow: instead of being able to use a with clause to handle all the possible things that can go wrong, we have to do some extra try/rescue wrapping to handle these exceptions.

Is there any appetite for making this behavior configurable? Or am I missing something here? It would be great to have a `start_link/1` setting that flipped connection errors to a tuple instead of an exception. With some guidance, I'd be happy to help make it happen.

josevalim:
I would love to see a PR exploring this and we can decide based on that if it is not too much work.

acco:
Ok @josevalim ! I'll see what I can whip up

jbavari:
@josevalim Would love any feedback you have on a PR I put together. Thank you in advance.

acco:
Per https://github.com/elixir-ecto/db_connection/pull/264#issuecomment-1161358040, think it makes the most sense to just funnel all calls to `Postgrex.query` into one location in our app and rescue the exception. Thanks all!

PR #264, opened 2022-06-21
Alter functions to pass error tuple instead of raising errors

jbavari:
As per #260 , we wanted a way to prevent raising errors and instead return error tuples.

I may be missing something - seeking some additional guidance to figure out what's needed to make it return error tuple instead of raising.

* Configuration - right now no clear pattern exists in `db_connection`. Either the configuration would have to be passed down to prevent raising, or have a simple `Application.get_env`, but keep sane defaults to raise. We do not want to disrupt [all consumers](https://hex.pm/packages?search=depends%3Ahexpm%3Adb_connection) of the `db_connection` library.
* All the consumers of `db_connection` would appear to need to adhere to raise vs error tuple, if indeed we did get this change in place.
* Tests around DBConnection (where the raise'ing happens) is very limited. My changes are quite fragile at the moment.
* Concerns around how to set up mocks/tests to get the raise/error tuples to trigger.
* By passing the tuples up, the call stack would receive unintended error tuples.
* An example Repo may have: `UserApplication.Repo -> Ecto.Repo -> Ecto.Adapters.Postgres -> Ecto.Repo.Schema -> Ecto.Adaptesr.SQL -> Ecto.Adapters.SQL.Connection -> Ecto.Adapters.Postgres.Connection -> DBConnection.execute`

Would love any feedback or suggestions the team would have. From where I sit, I would even think just overriding the userspace `Repo.insert` call to try/catch may be an easier fit over all. Happy to explore with the team!

### DIFF
--- a/.trunk/.gitignore
+++ b/.trunk/.gitignore
@@ -1,0 +1,4 @@
+*out
+*logs
+plugins
+user_trunk.yaml
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -1,0 +1,10 @@
+version: 0.1
+cli:
+  version: 0.16.1-beta
+lint:
+  enabled:
+    - actionlint@1.6.15
+    - git-diff-check@SYSTEM
+    - gitleaks@8.10.3
+    - markdownlint@0.32.1
+    - prettier@2.7.1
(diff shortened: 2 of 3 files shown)


josevalim:
Using the application environment is a no-no but you can pass this as an option. However, as much as it pains me to say this, this will definitely have further implications in transactions and similar, so the best is to most likely to rescue the exception. :(

PR #266, opened 2022-09-02
Add example implementation for a connection_listener

pdgonzalez872:
Hi!

As usual, thank you for the work in this repo! I noticed we didn't have a small example implementation for a `connection_listener`. After playing around with some code, I arrived at the the one below think it may serve as a nice starting point for folks wanting to implement their own `connection_listeners`.

The simplest way I found to test/prove it worked was to add it to a test, I do it [here](https://github.com/pdgonzalez872/db_connection/commit/845c2aaf8be97c7e1b65e335a4c73f13e74478f9)

Hope this helps!

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -435,44 +435,44 @@ defmodule DBConnection do
 
   Here is an example of a `connection_listener` implementation:
 
-    defmodule DBConnectionListener do
-      use GenServer
+      defmodule DBConnectionListener do
+        use GenServer
 
-      @impl true
-      def init(stack) when is_list(stack) do
-        {:ok, stack}
-      end
+        def start_link(opts) do
+          GenServer.start_link(__MODULE__, [], opts)
+        end
 
-      @impl true
-      def handle_call(:read_state, _from, state) do
-        {:reply, state, state}
-      end
+        def get_notifications(pid) do
+          GenServer.call(pid, :read_state)
+        end
 
-      @impl true
-      def handle_info({:connected, _pid} = msg, state) do
-        {:noreply, [msg | state]}
-      end
+        @impl true
+        def init(stack) when is_list(stack) do
+          {:ok, stack}
+        end
 
-      @impl true
-      def handle_info({_other_states, _pid} = msg, state) do
-        {:noreply, [msg | state]}
-      end
+        @impl true
+        def handle_call(:read_state, _from, state) do
+          {:reply, state, state}
+        end
 
-      def start_link(opts) do
-        GenServer.start_link(__MODULE__, [], opts)
-      end
+        @impl true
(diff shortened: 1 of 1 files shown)


lib/db_connection.ex
```diff
@@ -433,6 +433,38 @@ defmodule DBConnection do
...
   `pid` if you want to track all disconnections.
 
+  Here is an example of a `connection_listener` implementation:
+
+    defmodule DBConnectionListener do
+      use GenServer
+
+      def start_link() do
+        GenServer.start_link(__MODULE__, [], name: {:global, "db_connection_listener"})
+      end
```
v0idpwn, line 443: I'd suggest going for a more traditional `start_link/1` here, passing the options to `GenServer.start_link/3`. 

Instead of relying on name for `GenServer.call/3` you can use the pid. It would be nice if this was exported as a function instead of calling `GenServer.call/3` directly.

lib/db_connection.ex
```diff
@@ -433,6 +433,38 @@ defmodule DBConnection do
...
+
+      @impl true
+      def handle_call(:read_state, _from, state) do
+        {:reply, state, state}
+      end
+
+      @impl true
+      def handle_info(msg, state) do
+        {:noreply, [msg | state]}
+      end
```
v0idpwn, line 458: Can we perhaps make this smarter? For example, handling `:connected` and `:disconnected` messages differently?

v0idpwn (commented):
Really appreciate the PR, added some comments with ideas which (IMO) make it a nicer example.

lib/db_connection.ex
v0idpwn: Example: adding pids to the list on connect and removing them on disconnect

pdgonzalez872: Hi @v0idpwn! I actually went the opposite way and tried making it as dumb as possible. The goal was to have a  starting point on which folks could build on from.

lib/db_connection.ex
pdgonzalez872: > traditional start_link
Do you think it's better now?

> exported as a function
:white_check_mark: , good call

lib/db_connection.ex
pdgonzalez872: @v0idpwn, I added a function to match on the `:connected` state. Could you please expand on what you mean by "monitoring them and removing them on disconnect"? Sounds like a cool example! Please feel free to add commits to this if you'd like.

pdgonzalez872:
@v0idpwn sorry for taking so long. Pretty crazy couple of weeks. Thanks for the comments, they were good ones. I'm interested in the discussion [here](https://github.com/elixir-ecto/db_connection/pull/266#discussion_r973620245), I think that will make this an awesome example. Thanks again!

josevalim:

PR #271, opened 2022-11-22
Fix disconnect_all not working for intervals above 4294.

### LINKED ISSUE #270: `disconnect_all` does not work correctly with intervals above 4294
Please see https://github.com/whilefalse/disconnect_all for full reproduction information and more details.

## Brief overview

Calling `DBConnection.disconnect_all` with intervals above `4294` does not disconnect on checkin as expected, and as documented. With intervals above this value, the connection will only disconnect once it has become idle.

The reason for this is that we call `:erlang.phash2`, passing it the "erlang native time" representation of the interval. `:erlang.phash2` has a max input value of `2^32`, and `2^32` "erlang native time units" corresponds to `4294` milliseconds.

I'm aware that "native time units" can be OS dependent. I've tested this on my local development environment (macOS 12.6) and our staging environment (Alpine linux), and both return the same value for

```elixir
System.convert_time_unit(floor(:math.pow(2, 32)), :native, :millisecond)
```

i.e.

```
4294.
```

## Suggested fix

When calling `:erlang.phash2`, we don't really need native time resolution, we could just convert to milliseconds when calling `phash2`, then convert back to native time units.

I can put in a PR for this.

Thanks!

whilefalse:
fixes #270.

### DIFF
--- a/lib/db_connection/holder.ex
+++ b/lib/db_connection/holder.ex
@@ -252,7 +252,7 @@ defmodule DBConnection.Holder do
 
       true ->
         pid = :ets.lookup_element(holder, :conn, conn(:connection) + 1)
-        System.monotonic_time() > :erlang.phash2(pid, interval) + start
+        System.monotonic_time() > hash_pid(pid, interval) + start
     end
   rescue
     _ -> false
@@ -437,5 +437,11 @@ defmodule DBConnection.Holder do
 
   defp cancel_deadline(deadline) do
     :erlang.cancel_timer(deadline, async: true, info: false)
+  end
+
+  defp hash_pid(pid, interval_native) do
+    interval_ms = System.convert_time_unit(interval_native, :native, :millisecond)
+    hash = :erlang.phash2(pid, interval_ms)
+    System.convert_time_unit(hash, :millisecond, :native)
   end
 end


lib/db_connection/holder.ex
```diff
@@ -438,4 +438,10 @@ defmodule DBConnection.Holder do
   defp cancel_deadline(deadline) do
     :erlang.cancel_timer(deadline, async: true, info: false)
   end
+
+  defp hash_pid(pid, interval_native) do
+    interval_ms = System.convert_time_unit(interval_native, :native, :millisecond)
```
josevalim, line 443: What if we don't convert from milliseconds to native on `DBConnection.disconnect_all`? This way we can skip this initial conversion here.

whilefalse: Yep that's a good idea. I wasn't sure if it would be more confusing to have it stored in a different unit to the start timestamp, because I think it's stored as `{start, interval}`. But I can go ahead and do that if you think it makes more sense.

josevalim: Yeah, let's go ahead and change it there as well!

josevalim:

PR #275, opened 2023-02-10
Replace Connection with gen_statem

whatyouhide:
Just started on this but all tests pass 🙃 I also tried this branch out in ecto_sql and all tests pass there too. I'll see if I can try it in Postgrex maybe.

### DIFF
--- a/integration_test/ownership/manager_test.exs
+++ b/integration_test/ownership/manager_test.exs
@@ -575,7 +575,7 @@ defmodule ManagerTest do
       {:disconnect, RuntimeError.exception("oops"), TestConnection, :state}
     end
 
-    capture_log(fn ->
+    log = capture_log(fn ->
       assert Ownership.ownership_checkout(pool, pre_checkin: pre_checkin) == :ok
       assert_raise DBConnection.ConnectionError, "bad return value: :oops", fn ->
         assert_checked_out(pool, opts)
@@ -584,6 +584,9 @@ defmodule ManagerTest do
       assert_receive :pre_checkin
       assert_receive :reconnected
     end)
+
+    assert log =~ ~r"State machine #PID<\d+\.\d+\.\d+> terminating\n"
+    assert log =~ "** (RuntimeError) oops"
   end
 
   defp start_pool(opts \\ []) do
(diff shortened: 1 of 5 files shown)


josevalim:
@whatyouhide make sure that you run the integration tests in `ecto_sql`. :)

josevalim:
Looks fantastic. But note CI fails. :) And yeah, maybe it is simpler to not update `connection` indeed, so we keep it "stable".

lib/db_connection/connection.ex
```diff
@@ -87,16 +95,28 @@ defmodule DBConnection.Connection do
...
+          fn ->
+            [
+              inspect(mod),
+              " (",
+              inspect(self()),
+              ") failed to connect: " | Exception.format_banner(:error, err, [])
+            ]
+          end,
+          crash_reason: {err, []}
+        )
```
whatyouhide, line 118: @josevalim I had to add this here too. From testing with ecto_sql, it turns out that `raise`ing here doesn't log anything for gen_statem, but it did log for Connection. I’m confused, but I also suck at OTP logging, so I’m not surprised 🙃 

If you have better ideas, I’m all ears. 🙃

integration_test/ownership/manager_test.exs
```diff
@@ -575,15 +575,15 @@ defmodule ManagerTest do
       {:disconnect, RuntimeError.exception("oops"), TestConnection, :state}
     end
 
-    assert capture_log(fn ->
+    capture_log(fn ->
```
josevalim, line 578: Perhaps we should keep the assertion and update it just in case? Especially as you said they report errors different?

whatyouhide: @josevalim yeah, I had to add a manual log again. I’m not sure why gen_statem is not logging anything when stopping with reason other than normal/shutdown 😵‍💫

josevalim:
@whatyouhide can you check CI? is_exception is not available, you will have to pattern match. :)

whatyouhide:
Oooph yeah, I stopped checking because it was giving trouble with Erlang/Elixir versions yesterday, and I thought it was still that. Fixed now.

josevalim:

PR #280, opened 2023-03-16
Add DBConnection.Pool behaviour

whatyouhide:
We can keep this private for now, but I think it would be nice to have it internally and possibly expose it one day?


josevalim:

PR #281, opened 2023-03-17
Improve exceptions docs in DBConnection.run/3

whatyouhide:


lib/db_connection.ex
```diff
@@ -793,7 +798,14 @@ defmodule DBConnection do
   `transaction/3` call inside another `transaction/3` will be treated
   the same as `run/3`.
 
-  ### Options
+  > #### Checkout failures {: .warning}
+  >
+  > If we cannot check out a connection from the pool, this function raises a
+  > `DBConnection.ConnectionError` exception. This could also happen in some
+  > other cases, so if you want to handle this case, you should rescue
+  > `DBConnection.ConnectionError` exceptions when using `run/3`.
```
josevalim, line 806: "This could also happen in some other cases". This is a bit unclear. This could also happen? This means what here: raising? and which other cases? :)

whatyouhide: @josevalim it's worse than I thought, because we raise a `DBConnection.ConnectionError` even for bad return values from callbacks. I’m thinking for those it would be better to have a different error, but for now I amended the PR with a bit better docs.

wojtekmach: bad return sounds like an ArgumentError (kind of the opposite tho) and I think separate exception makes sense as these are always bugs.

lib/db_connection.ex
```diff
@@ -793,7 +798,16 @@ defmodule DBConnection do
...
 
-  ### Options
+  > #### Checkout failures {: .warning}
+  >
+  > If we cannot check out a connection from the pool, this function raises a
+  > `DBConnection.ConnectionError` exception. This function also raises a
+  > `DBConnection.ConnectionError` exception if a `DBConnection` callback
+  > invoked because of code in `fun` returns a bad value.
+  > If you want to handle these cases, you should rescue
+  > `DBConnection.ConnectionError` exceptions when using `run/3`.
```
josevalim, line 808: Two additional comments:

1. The invalid return type means an adapter is buggy, so it is always a bug upstream which should be addressed (but in theory should never happen), so I am not sure if we should mention it

2. "If you want to handle these cases" - I assume the only case of interest is "if we cannot check out a connection from the pool", right?

So my suggestion is:

```suggestion
  > `DBConnection.ConnectionError` exception.
  > If you want to handle these cases, you should rescue
  > `DBConnection.ConnectionError` exceptions when using `run/3`.
```

In other words, sure, other exceptions can happen, but they should not happen unless a bug outside of our code (and outside of the usual logic of not being able to checkout).

whatyouhide:
@josevalim sounds good, updated!

josevalim:

PR #282, opened 2023-03-19
Raise RuntimeError on bad return values

whatyouhide:
As mentioned in #281. cc @wojtekmach, I went with `RuntimeError`, which I think makes more sense here than `ArgumentError`.

### DIFF
--- a/integration_test/cases/after_connect_test.exs
+++ b/integration_test/cases/after_connect_test.exs
@@ -174,7 +174,7 @@ defmodule AfterConnectTest do
            ] = A.record(agent)
   end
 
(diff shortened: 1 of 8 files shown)


josevalim:
On the off chance someone is relying on this, I don’t think this is worth changing, given how minimal it is. :)

josevalim:
Isn't this propagated upstream to `run/3` as per the previous PR?

whatyouhide:
Yes, but I mean, you think they're relying on possibly returning bad values (like in Postgrex or something)? I think that would be considered a bug, so they should probably not be catching it?

josevalim:
I don’t think they are or should be, but this is minor to justify potentially breaking something. That’s all :)

PR #290, opened 2023-07-27
Add DBConnection.available_start_options/0

whatyouhide:
This would be awesome in [Xandra](https://github.com/lexhide/xandra) for example, because we could split start options in

1. Xandra options, which we validate
2. DBConnection options, which we pass down to DBConnection
3. Any other option, for which we'd raise as "unknown option"

If this is a good addition, I can do the same for `t:DBConnection.option/0`.


josevalim:

PR #291, opened 2023-07-27
Add DBConnection.available_connection_options/0

whatyouhide:
Follow up to #290.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -580,11 +580,11 @@ defmodule DBConnection do
   `DBConnection`, since they can use the return value of this function to perform
   validation on options only passing down these options to DBConnection.
 
-  See also `t:option/0`.
+  See also `t:connection_option/0`.
   """
   @doc since: "2.6.0"
-  @spec available_options() :: [atom, ...]
-  def available_options do
+  @spec available_connection_options() :: [atom, ...]
+  def available_connection_options do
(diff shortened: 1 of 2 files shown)


josevalim:
`available_connection_options` perhaps?

josevalim:
It seems most of our specs use `Keyword.t()`? Should we align that instead then? But if we are doing that, I would prefer to call it `t:DBConnection.connection_options` to avoid ambiguity?

whatyouhide:
@josevalim yeah, the type is there but not really used. If you want, we can add `t:DBConnection.connection_option/0`, deprecate `t:DBConnection.option/0`, and change specs to use `DBConnection.connection_option/0`. Thoughts?

josevalim:
Sure!

lib/db_connection.ex
```diff
@@ -134,12 +134,30 @@ defmodule DBConnection do
...
+  > #### Deprecated {: .warning}
+  >
+  > This option is deprecated since v2.6.0. Use `t:connection_option/0` instead.
+
+  """
   @type option ::
           {:log, (DBConnection.LogEntry.t() -> any) | {module, atom, [any]} | nil}
           | {:queue, boolean}
           | {:timeout, timeout}
           | {:deadline, integer | nil}
```
josevalim, line 149:
```suggestion
  @type option :: connection_option
```

josevalim:

PR #292, opened 2023-08-10
Add metrics for connection pool usage

peixian:
This PR adds support for a `DBConnection.ConnectionPool.get_metrics()` function that returns the current state of active and waiting connections, to allow upstream callers to gain some visibility on their database usage. We're using a form of this in our internal forks.

### DIFF
--- a/lib/db_connection/connection_pool.ex
+++ b/lib/db_connection/connection_pool.ex
@@ -17,6 +17,8 @@ defmodule DBConnection.ConnectionPool do
   @queue_interval 1000
   @idle_interval 1000
   @time_unit 1000
+  @active_counter_idx 1
+  @waiting_counter_idx 2
 
   @doc false
   def start_link({mod, opts}) do
@@ -60,6 +62,7 @@ defmodule DBConnection.ConnectionPool do
     interval = Keyword.get(opts, :queue_interval, @queue_interval)
     idle_interval = Keyword.get(opts, :idle_interval, @idle_interval)
     idle_limit = Keyword.get_lazy(opts, :idle_limit, fn -> Keyword.get(opts, :pool_size, 1) end)
+    pool_size = Keyword.get(opts, :pool_size, 1)
     now_in_native = System.monotonic_time()
     now_in_ms = System.convert_time_unit(now_in_native, :native, @time_unit)
 
@@ -75,10 +78,9 @@ defmodule DBConnection.ConnectionPool do
       idle: nil
     }
 
-    metrics = %{
-      active: 0,
-      waiting: 0
-    }
+    metrics = :counters.new(2, [])
+    # when the pool starts up, it checks out and should decrement this to 0
+    :counters.put(metrics, @active_counter_idx, pool_size)
 
     codel = start_idle(now_in_native, start_poll(now_in_ms, now_in_ms, codel))
     {:ok, {:busy, queue, codel, ts, metrics}}
@@ -90,6 +92,12 @@ defmodule DBConnection.ConnectionPool do
     {:reply, :ok, {type, queue, codel, ts, metrics}}
   end
 
+  def handle_call(:get_metrics, _from, {_, _, _, _, metrics} = state) do
+    active = :counters.get(metrics, @active_counter_idx)
+    waiting = :counters.get(metrics, @waiting_counter_idx)
+    {:reply, %{active: active, waiting: waiting}, state}
+  end
+
   @impl GenServer
   def handle_info(
         {:db_connection, from, {:checkout, _caller, now, queue?}},
@@ -98,7 +106,8 @@ defmodule DBConnection.ConnectionPool do
     case queue? do
       true ->
         :ets.insert(queue, {{now, System.unique_integer(), from}})
-        {:noreply, {:busy, queue, codel, ts, %{metrics | waiting: metrics.waiting + 1}}}
+        :counters.add(metrics, @waiting_counter_idx, 1)
+        {:noreply, {:busy, queue, codel, ts, metrics}}
 
       false ->
         message = "connection not available and queuing is disabled"
@@ -115,7 +124,8 @@ defmodule DBConnection.ConnectionPool do
     case :ets.first(queue) do
       {queued_in_native, holder} = key ->
         Holder.handle_checkout(holder, from, queue, queued_in_native) and :ets.delete(queue, key)
-        {:noreply, {:ready, queue, codel, ts, %{metrics | active: metrics.active + 1}}}
+        :counters.add(metrics, @active_counter_idx, 1)
+        {:noreply, {:ready, queue, codel, ts, metrics}}
 
       :"$end_of_table" ->
         handle_info(checkout, put_elem(ready, 0, :busy))
@@ -210,6 +220,7 @@ defmodule DBConnection.ConnectionPool do
          {queued_in_native, holder} = key when queued_in_native <= past_in_native <-
            :ets.first(queue) do
       :ets.delete(queue, key)
+      :counters.sub(metrics, @waiting_counter_idx, 1)
       Holder.maybe_disconnect(holder, elem(ts, 0), 0) or Holder.handle_ping(holder)
       drop_idle(past_in_native, limit - 1, status, queue, codel, ts, metrics)
     else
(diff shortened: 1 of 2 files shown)


lib/db_connection/connection_pool.ex
```diff
@@ -358,4 +372,9 @@ defmodule DBConnection.ConnectionPool do
     idle = :erlang.start_timer(timeout, self(), now_in_native, abs: true)
     %{codel | idle: idle}
   end
+
+  defp decrement_or_zero_active_conn(metrics) do
```
josevalim, line 376: Why do we need this logic? If at some point we can go to -1, it means at some point we were off by one?

josevalim:
Hi @peixian! This looks good to me but I would consider using `:counters` instead, so we can read them decentralized. We would also need tests. :)

whatyouhide:
`counters` is such a good idea for this 👏

lib/db_connection/connection_pool.ex
peixian: Our main usage of this was Xandra, and we noticed that when Xandra started up the connection pool, it seems to checkin the connections, so a simple -1 each time would cause your pool metrics to startup as `{-N, 0}` and the highest point it would reach was 0.

peixian:
@josevalim I'll change it to counters. For tests, it appears that the tests are by default already broken? I didn't add onto it because I wasn't sure what the state of tests are.

For example:

```
 ~/c/db_connection_master (master)> git log | head                                                                                             
commit 022393b7895f5bd3a7e884a35c7590f4fa549986
Author: Andrea Leopardi <an.leopardi@gmail.com>
Date:   Thu Jul 27 18:33:49 2023 +0200

    Add DBConnection.available_connection_options/0 (#291)
    
    
~/c/db_connection_master (master) [2]> mix test .     
.......
Finished in 1.7 seconds (1.7s async, 0.00s sync)
202 tests, 32 failures
    
```

josevalim:
Instead of `mix test .`, please do `mix test.all`.

Also, when a new connection is started, it always checks itself in once ready to use, so you need to take that into account.

peixian:
@josevalim this is good for re-review, I've got all the tests passing and added a few.

lib/db_connection/connection_pool.ex
```diff
@@ -40,6 +42,13 @@ defmodule DBConnection.ConnectionPool do
     GenServer.call(pool, {:disconnect_all, interval}, :infinity)
   end
 
+  @doc """
+  Returns connection metrics in the shape of %{active: N, waiting: N}
+  """
+  def get_connection_metrics(pid) do
+    GenServer.call(pid, :get_metrics)
+  end
+
```
v0idpwn, line 51: Now that you're using counters, you don't need to call the connection pool, you can read them directly on this function

lib/db_connection/connection_pool.ex
```diff
@@ -68,25 +78,36 @@ defmodule DBConnection.ConnectionPool do
...
+  def handle_call({:disconnect_all, interval}, _from, {type, queue, codel, _ts, metrics}) do
     ts = {System.monotonic_time(), interval}
-    {:reply, :ok, {type, queue, codel, ts}}
+    {:reply, :ok, {type, queue, codel, ts, metrics}}
+  end
+
+  def handle_call(:get_metrics, _from, {_, _, _, _, metrics} = state) do
+    active = :counters.get(metrics, @active_counter_idx)
+    waiting = :counters.get(metrics, @waiting_counter_idx)
+    {:reply, %{active: active, waiting: waiting}, state}
```
josevalim, line 98: You can return the counter to the client now, and the client reads it. :)

```suggestion
    {:reply, metrics, state}
```

lib/db_connection/connection_pool.ex
```diff
@@ -68,25 +78,36 @@ defmodule DBConnection.ConnectionPool do
       idle: nil
     }
 
+    metrics = :counters.new(2, [])
+    # when the pool starts up, it checks out and should decrement this to 0
```
josevalim, line 82:
```suggestion
    # as the pool starts, connections are checked in and should decrement this to 0
```

test/db_connection/metrics_test.exs
```diff
@@ -0,0 +1,126 @@
+defmodule MetricsTest do
```
josevalim, line 1:
```suggestion
defmodule DBConnection.MetricsTest do
```

test/db_connection/metrics_test.exs
```diff
@@ -0,0 +1,126 @@
...
+  use ExUnit.Case, async: false
+
+  defmodule DummyDB do
```
josevalim, line 4: We already have a DummyDB that we use in other tests (via the agent+connection). Could we perhaps re-use that instead?

test/db_connection/metrics_test.exs
```diff
@@ -0,0 +1,126 @@
...
+    defdelegate handle_declare(arg1, arg2, arg3, arg4), to: DummyDB
+    defdelegate handle_fetch(arg1, arg2, arg3, arg4), to: DummyDB
+    defdelegate handle_prepare(arg1, arg2, arg3), to: DummyDB
+    defdelegate handle_rollback(arg1, arg2), to: DummyDB
+    defdelegate handle_status(arg1, arg2), to: DummyDB
+    defdelegate handle_execute(arg1, arg2, arg3, arg4), to: DummyDB
+    defdelegate ping(arg1), to: DummyDB
+
+    def checkout(state) do
+      :timer.sleep(1000)  # Simulate a delay
```
josevalim, line 48: Those tests are going to be very expensive, unfortunately. It may be this single test file is slower than the whole test suite. :( Would it be possible for us to use :connection_listeners to detect when all connections are up instead of relying on long :timer.sleep periods?

lib/db_connection/connection_pool.ex
peixian: I considered that, but two big issues came up:

1. This makes the call not `:rpc` safe, since the counters returns an atomic reference that's local, and it's quite likely that you'd attempt to call this on a different node, in the situation where you have a production application on one node and a manager on another node that might be polling for metrics. (I have some testing code in this comment below for this. 
2. Passing back the counter without unwrapping imo is a bad user experience. It's not super clear which value in the counter references which. 

----------
Testing counters:
```
# node 1 iex --name node1@127.0.0.1         
defmodule CounterServer do
  use GenServer

  def start_link do
    GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
  end

  def get_counter do
    GenServer.call(__MODULE__, :get)
  end


  def init(:ok) do
    # Initialize an Erlang counter
    counter = :counters.new(1, [])
    {:ok, counter}
  end

  def handle_call(:get, _from, counter) do
    # Fetch the value from the counter
    {:reply, counter, counter}
  end
end


{:ok, pid} = CounterServer.start_link(0)
```

```
# node 2 iex --name node2@127.0.0.1         

Node.connect(:"node1@127.0.0.1")

# You can define a lightweight module to delegate to the GenServer on node1
defmodule CounterClient do
  def get_counter_on_node1 do
    :rpc.call(:"node1@127.0.0.1", CounterServer, :get_counter, [])
  end
end

counter = CounterClient.get_counter_on_node1()

 :counters.get(counter, 0)
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: invalid atomics reference
  * 2nd argument: out of range

    :counters.get({:atomics, #Reference<12104.2198675303.198049794.112940>}, 0)
    iex:5: (file)
```

test/db_connection/metrics_test.exs
peixian: Hm, I figured I'd just reimplement a smaller DummyDB here for maximum clarity and to segregate the tests from the other ones.

test/db_connection/metrics_test.exs
peixian: Hm, I'm not sure, I'll take a look

lib/db_connection/connection_pool.ex
greg-rychlewski: I think instead of returning the counter directly from one node to the other you could instead create a function that:

1. Knows what the different values mean
2. Can read the counter and pass the values nicely in a map or something like that to the other node

The alternative of letting something continuously poll the connection pool for metrics is not ideal imo.

test/db_connection/metrics_test.exs
v0idpwn: Alternatively, regular erlang tracing can be used.

lib/db_connection/connection_pool.ex
josevalim: Those are good points. Reading the counters are quick anyway, so we can keep this as is. :)

PR #293, opened 2023-08-11
Add :connection_listeners_tag

whatyouhide:
As discussed with @josevalim.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -119,6 +119,7 @@ defmodule DBConnection do
           {:after_connect, (t -> any) | {module, atom, [any]} | nil}
           | {:after_connect_timeout, timeout}
           | {:connection_listeners, list(Process.dest()) | nil}
+          | {:connection_listeners_tag, any}
           | {:backoff_max, non_neg_integer}
           | {:backoff_min, non_neg_integer}
           | {:backoff_type, :stop | :exp | :rand | :rand_exp}
@@ -391,6 +392,9 @@ defmodule DBConnection do
     * `:connection_listeners` - A list of process destinations to send
       notification messages whenever a connection is connected or disconnected.
       See "Connection listeners" below
+    * `:connection_listeners_tag` - A term that is sent alongside
+      `:connected`/`:disconnected` messages to connection listeners.
+      See "Connection listeners" below. Available since v2.6.0
     * `:name` - A name to register the started process (see the `:name` option
       in `GenServer.start_link/3`)
     * `:pool` - Chooses the pool to be started (default: `DBConnection.ConnectionPool`). See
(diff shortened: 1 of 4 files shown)


lib/db_connection/connection.ex
```diff
@@ -75,6 +75,7 @@ defmodule DBConnection.Connection do
       timer: nil,
       backoff: Backoff.new(opts),
       connection_listeners: Keyword.get(opts, :connection_listeners, []),
+      connection_listeners_tag: Keyword.fetch(opts, :connection_listeners_tag),
```
josevalim, line 78: Let's please keep them as a single field since they are always read together!

lib/db_connection.ex
```diff
@@ -454,12 +458,14 @@ defmodule DBConnection do
 
     * `{:connected, pid}`
     * `{:disconnected, pid}`
+    * `{:connected, pid, term}` if `:connection_listeners_tag` is set to `term`
+    * `{:disconnected, pid, term}` if `:connection_listeners_tag` is set to `term`
```
josevalim, line 462:
```suggestion
    * `{:connected, pid, term}` if `:connection_listeners_tag` is set
    * `{:disconnected, pid, term}` if `:connection_listeners_tag` is set
```

lib/db_connection/connection.ex
whatyouhide: @josevalim what do you have in mind?

  1. `:connection_listeners_with_tag`, which becomes exclusive with `:connection_listeners`
  2. `connection_listeners: [...] | {[...], tag}`
  3. ?

🙃

josevalim: `connection_listeners: [...] | {[...], tag}` or `connection_listeners: {[...], {:ok, tag} | :error}}`

whatyouhide:
@josevalim updated!

lib/db_connection/connection.ex
```diff
@@ -506,9 +506,18 @@ defmodule DBConnection.Connection do
...
+  defp notify_connection_listeners(action, %{} = state) do
     %{connection_listeners: connection_listeners} = state
 
-    Enum.each(connection_listeners, &send(&1, message))
+    {listeners, message} =
+      case connection_listeners do
+        listeners when is_list(listeners) and action in [:connected, :disconnected] ->
+          {listeners, {action, self()}}
+
+        {listeners, tag} when is_list(listeners) and action in [:connected, :disconnected] ->
```
josevalim, line 517:
```suggestion
        listeners when is_list(listeners) ->
          {listeners, {action, self()}}

        {listeners, tag} when is_list(listeners) ->
```

josevalim:

Issue #296, opened 2023-09-03
Allow drivers to return customized `begin` queries.


greg-rychlewski:
**Problem Statement**

I'm looking to add ddl transactions to SQLite. To prevent concurrent migrations I'm trying to use the exclusive transaction shown here: https://www.sqlite.org/lang_transaction.html#deferred_immediate_and_exclusive_transactions

As part of this it would be nice to log/test that the right statements are being executed by the database. But right now db connection will log a generic `begin` statement because the result from `handle_begin` is `{:ok, result, state}`.

**Proposal**

Allow `handle_begin` callbacks to return `{:ok, query, result, state}` in addition to `{:ok, result, state}` for drivers that can use customized begin queries. This would mirror the way `handle_execute` enables logging.

v0idpwn:
I don't see why not.

josevalim:
Sounds good to me!

PR #297, opened 2023-09-04
Allow handle_begin to return query

greg-rychlewski:
Closes https://github.com/elixir-ecto/db_connection/issues/296

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -201,9 +201,12 @@ defmodule DBConnection do
   @doc """
   Handle the beginning of a transaction.
 
-  Return `{:ok, result, state}` to continue, `{status, state}` to notify caller
-  that the transaction can not begin due to the transaction status `status`,
-  or `{:disconnect, exception, state}` to error and disconnect.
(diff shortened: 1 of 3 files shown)


lib/db_connection.ex
```diff
@@ -1775,6 +1788,9 @@ defmodule DBConnection do
       {status, _conn_state} when status in [:idle, :transaction, :error] ->
         status_disconnect(conn, status, meter)
 
+      {:ok, query, result, _conn_status} ->
+        {:ok, {query, result}, meter}
```
greg-rychlewski, line 1792: I packed them into a tuple because if we return `{:ok, query, result, meter}` here then we have to handle the following 4 return types:

- `{:ok, result, meter}`
- `{:ok, query, result, meter}`
- `{:ok, conn, result, meter}`
- `{:ok, conn, query, result, meter}` 

The first 2 are when you start a transaction inside of `run`. The other 2 when you do a fresh checkout. 

It seems less complicated to pack these into a tuple but please let me know what you think.

v0idpwn (approved):

josevalim (approved):

PR #301, opened 2023-11-22
Fix `pool_size` should be equal or greater to 1

Dvogiatz:
Hello, I've noticed that you are allowed to pass 0 or negative pool_size, which then creates connections instead of raising.


lib/db_connection/connection_pool/pool.ex
```diff
@@ -16,6 +16,9 @@ defmodule DBConnection.ConnectionPool.Pool do
   @impl true
   def init({owner, tag, mod, opts}) do
     size = Keyword.get(opts, :pool_size, 1)
+
+    if size < 1, do: raise(ArgumentError, "Pool size must be greater or equal to 1. Got #{size}.")
```
josevalim, line 20:
```suggestion
    if size < 1, do: raise(ArgumentError, "pool size must be greater or equal to 1, got #{size}")
```

josevalim:

PR #304, opened 2024-01-05
Add DBConnection.ConnectionError.t/0 type

whatyouhide:
I also moved the connection error definition into its own file and added some guards to its `exception/2` function.

The idea here is to lay the foundations for expanding `:reason` so that programmers can gather more from that instead of sometimes having to match on the message.


josevalim:

PR #306, opened 2024-03-20
Add metrics for connection pool usage (tests speed up + checkin metrics bugfix)

kor-and-m:


kor-and-m:
Hello @josevalim, I highly motivated to complete the pull request could you take a look at it please?

josevalim:
Thank you for the PR.

Shouldn't be able to get these metrics from querying the ETS table? We are either with connections in the pools or with pending requests, the mode tells if we are busy or ready. :)

Also, could we reuse the existing testing infrastructure somehow? Or do we need to define new DBs and new Pools?

PR #307, opened 2024-05-21
Add metrics for connection pool usage v3

realglebivanov:
Hello!

Since these two PRs seem to be abandoned

https://github.com/elixir-ecto/db_connection/pull/306
https://github.com/elixir-ecto/db_connection/pull/292,

I've decided to open another one that will hopefully be merged.


lib/db_connection/connection_pool.ex
```diff
@@ -358,4 +378,23 @@ defmodule DBConnection.ConnectionPool do
     idle = :erlang.start_timer(timeout, self(), now_in_native, abs: true)
     %{codel | idle: idle}
   end
+
+  defp execute_queue_telemetry(status, queue, event, count \\ 1)
```
josevalim, line 382: I am not the biggest fun of executing those events and traversing the table on every operation, this can be quite expensive in the long term. I think a pull design will probably be a better fit here, as commented in the other PR. Basically, have an operation that reads the ETS the status and the ETS table on demand.

realglebivanov: Though this approach is quite dubious, there are a couple of thoughts that I've had:

1. It's a clearer API in comparison to finding a pool PID and then calling a function from an internal module
2. If the pool is ready, then there are either not that many queries to perform or checked in connection count is pretty low. In both cases an ets table scan is probably acceptable.
3. In case the pool is busy and the queue length is huge, the problem would be probably not the `ets.select_count` call(or it would be a much lesser problem). If the queue length is small, an ets scan wouldn't seem to be a problem.
4. Also, the approach with telemetry calls is easier to test

@josevalim What is your opinion?

josevalim:
The pool ID should be the name of the Ecto repo, no? My biggest concern is that those costs will be paid by everyone in every operation, regardless of the sampling or need. So I think we should try polling first and fallback to telemetry only if that doesn’t work. Or put the telemetry in this PR behind some periodic action.

realglebivanov:
@josevalim 
Hey, thanks for the review!
I fixed the issue you have found but the tests are a little bit ugly.
Hopefully, we'll find the correct solution.

josevalim:
Thank you, this looks great! I think we just need to move it to the `DBConnection` module now, which is the public API. But perhaps, to do so, we need a callback in the pool and implement it in the ownership pool too? 🤔

realglebivanov:
Fair enough but this will introduce a breaking change into the API. 
Every custom pool there is will have to implement the new callback.
Also, it would probably be better to move this to DBConnection.Pool as it seems to be a part of the API as well

realglebivanov:
Actually, it isn't a part of the public API but I guess it should be a part of both DBConnection.Pool and DBConnection.

josevalim:
Yes, the pool API is private, so we are fine (but it would be fine to add new callbacks to behaviours nonetheless).

realglebivanov:
@josevalim I've managed to come up with a sketch of what the public API might look like. Take a look please!

lib/db_connection/ownership/manager.ex
```diff
@@ -98,6 +104,38 @@ defmodule DBConnection.Ownership.Manager do
...
+      |> Enum.map(fn {_, {proxy, _, _}} ->
+        if Process.alive?(proxy) do
+          GenServer.call(proxy, :get_connection_metrics)
+        end
+      end)
+      |> Enum.reject(&is_nil/1)
+
+    {:reply, {:ok, pool_metrics ++ proxy_metrics}, state}
+  rescue
+    error in MatchError ->
```
josevalim, line 121: When would this happen?

lib/db_connection/ownership/manager.ex
```diff
@@ -98,6 +104,38 @@ defmodule DBConnection.Ownership.Manager do
...
 
   @impl true
+  def handle_call(:get_connection_metrics, _from, %{pool: pool, owners: owners, log: log} = state) do
+    {:ok, pool_metrics} = DBConnection.ConnectionPool.get_connection_metrics(pool)
+
+    proxy_metrics =
+      owners
+      |> Enum.map(fn {_, {proxy, _, _}} ->
+        if Process.alive?(proxy) do
+          GenServer.call(proxy, :get_connection_metrics)
```
josevalim, line 114: Process.alive? Can always be a race because the process can die immediately after, so I would rather move the try/catch up here and catch any exit. :)

josevalim (commented):
This looks great, beautifully done! Just one last question: do you prefer to return a list of metrics (as in the PR) or return a single metric map where we sum (and/or average) the metrics from the ownership pool?

lib/db_connection/ownership/manager.ex
realglebivanov: It literally wouldn't. It was meant to hightlight the return type of `ConnectionPool.get_connection_metrics`

josevalim: So we can remove this then :)

lib/db_connection/ownership/manager.ex
realglebivanov: I thought it would reduce the probability of race condition a little bit. 
I don't think it's possible to move `catch :exit, _` here because `DBConnection.ConnectionPool.get_connection_metrics` call may cause exit(e.g. due to timeout) too

lib/db_connection/ownership/manager.ex
realglebivanov: Deleted it since it causes confusion

realglebivanov:
Thanks, José! 
I'd rather leave it as it is if you don't mind.
I just wanted the end user to decide in which way it would be convenient to aggregate the metrics.

lib/db_connection/ownership/manager.ex
josevalim: exactly, hence the suggestion of wrapping exactly that call.

lib/db_connection/ownership/manager.ex
```diff
@@ -98,6 +104,31 @@ defmodule DBConnection.Ownership.Manager do
...
+  catch
+    :exit, reason ->
+      if log do
+        Logger.log(
+          log,
+          "Caught :exit while calling :get_connection_metrics due to #{inspect(reason)}"
+        )
+      end
+
+      {:reply, :error, state}
```
josevalim, line 129:
```suggestion
    {:ok, pool_metrics} = DBConnection.ConnectionPool.get_connection_metrics(pool)

    proxy_metrics =
      owners
      |> Enum.map(fn {_, {proxy, _, _}} ->
        try do
          GenServer.call(proxy, :get_connection_metrics)
        catch
          :exit, _ when log -> 
            if log do
              Logger.log(
                log,
                "Caught :exit while calling :get_connection_metrics due to #{inspect(reason)}"
              )
            end

            nil
        end
      end)
      |> Enum.reject(&is_nil/1)

    {:reply, {:ok, pool_metrics ++ proxy_metrics}, state}
```

lib/db_connection/ownership/manager.ex
realglebivanov: We would have to wrap both `ConnectionPool.get_connection_metrics` and `GenServer.call(proxy, :get_connection_metrics)` in that case. 
Wouldn't it be easier to catch all the exits in one place? 
Am I getting something wrong?

lib/db_connection/ownership/manager.ex
josevalim: Here is the suggestion. The poll is linked to the current process, so we don't have to worry about it crashing, cause it would crash us too. There is a chance of timeout, but we can also just pass `:infinity` as the timeout (which I recommend anyway).

The only pending question is: do you want to return `{:ok, metrics}` or just the metrics directly?

lib/db_connection/connection_pool.ex
```diff
@@ -40,6 +40,12 @@ defmodule DBConnection.ConnectionPool do
     GenServer.call(pool, {:disconnect_all, interval}, :infinity)
   end
 
+  @doc false
+  @impl DBConnection.Pool
+  def get_connection_metrics(pool) do
+    GenServer.call(pool, :get_connection_metrics)
```
josevalim, line 46:
```suggestion
    GenServer.call(pool, :get_connection_metrics, :infinity)
```

lib/db_connection/ownership/manager.ex
```diff
@@ -59,6 +59,12 @@ defmodule DBConnection.Ownership.Manager do
     GenServer.call(manager, {:allow, parent, allow}, timeout)
   end
 
+  @spec get_connection_metrics(GenServer.server()) ::
+          {:ok, [DBConnection.Pool.connection_metrics()]} | :error
+  def get_connection_metrics(manager) do
+    GenServer.call(manager, :get_connection_metrics)
```
josevalim, line 65:
```suggestion
    GenServer.call(manager, :get_connection_metrics, :infinity)
```

lib/db_connection/ownership/manager.ex
josevalim: We don't need to wrap the pool one. And wapping the second one means we can provide results even if one of the pools is failing.

lib/db_connection/ownership/manager.ex
realglebivanov: We can't return just metrics because it's the part of DBConnection.Pool behavior.
By the way, do we really want the client of get_connection_metrics to wait indefinitely for a response or handle timeouts on its own?

realglebivanov: Actually, we've gotten rid of potential errors. We probably get rid of maybeish type in the public API as well

josevalim: > By the way, do we really want the client of get_connection_metrics to wait indefinitely for a response

Why would we wait indefinitely? A timeout is a safe default if we have blocking code but that's not the case here. The whole pool is design to queue things immediately instead of blocking.

realglebivanov: I meant the `GenServer.call` calls with `timeout = :infinity` which in theory could cause indefinite wait time

josevalim:

realglebivanov:
:heart: i'm a contributor now :rofl:

josevalim:
And not an easy feature to add, well done!

PR #310, opened 2024-06-23
Add a telemetry event on connect/disconnect

v0idpwn:
Adds a telemetry event on connection/disconnection.

Moves the logic of notifying listeners to a telemetry handler.


josevalim (changes requested):
Thanks but I think prefer the opposite route. Listeners are free, telemetry has a tiny cost. Can we have a built in listeners that pushes telemetry events?

v0idpwn:
Sounds good, will send another PR with the alternative approach :)

v0idpwn:
Closing this in favor of #311

PR #311, opened 2024-06-24
Add a connection listener that emits telemetry events

v0idpwn:
Alternative to #310. Would have to write documentation, etc.

### DIFF
--- a/lib/db_connection/telemetry_listener.ex
+++ b/lib/db_connection/telemetry_listener.ex
@@ -1,16 +1,49 @@
 defmodule DbConnection.TelemetryListener do
   @moduledoc """
   A connection listener that emits telemetry events for connection and disconnection
+
+  It monitors connection processes and ensures that disconnection events are
+  always emitted.
+
+  ## Telemetry events
+
+  ### Connected
+
+  `[:db_connection, :connected]` - Executed after a connection is established.
+
+  #### Measurements
+
+    * `:count` - Always 1
+
+  #### Metadata
+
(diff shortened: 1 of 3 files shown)


josevalim (commented):
I like this! We need to improve the docs to say exactly how to use and the events emitted but it works for me!

v0idpwn:
I added two documentation sections. I'm afraid it's a bit hidden in the documentation, but the `DBConnection` module doc is already quite big :)

lib/db_connection/telemetry_listener.ex
```diff
@@ -0,0 +1,101 @@
...
+  ## Usage
+
+  Start the listener, optionally using a name, and pass it under the
+  `:connection_listeners` option when starting DbConnection:
+
+      {:ok, pid} = TelemetryListener.start_link([name: MyListener])
+      {:ok, _conn} = DBConnection.start_link(SomeModule, [connection_listeners: [MyListener]])
+      # Using a tag, which will be sent in telemetry metadata
+      {:ok, _conn} = DBConnection.start_link(SomeModule, [connection_listeners: {[pid], :my_tag}])
+
```
josevalim, line 17: What do you think about showing an example with a supervision tree perhaps?

josevalim (approved):
One tiny comment and ship it!

v0idpwn:
Almost merged it as `DbConnection` instead of `DBConnection` 😄

josevalim:
Done!

PR #314, opened 2024-09-29
Enable descentralized counters

josevalim:


josevalim:

greg-rychlewski:
@josevalim Sorry to bug you. I read the article about decentralized counters after seeing this PR but I'm not 100% I understand what they are used for. Is this some kind of internal counter each ets table keeps to know how many times it has been updated?

josevalim:
It is a counter storing how many entries in the table. But we never query its size. :)

josevalim:
Also, you asking questions is never bugging me, I am always glad to transfer knowledge on anything Elixir/Ecto related :)

Issue #315, opened 2024-11-09
Guidance on handling closed connections


ruslandoga:
👋 

I wonder if it's possible to gracefully reconnect once the underlying adapter's connection is detected to be closed to avoid `:closed` errors on user queries (DBConnection.query / DBConnection.execute)? For example, maybe we could use [`:inet.monitor/1`](https://www.erlang.org/doc/apps/kernel/inet.html#monitor/1) from the adapters? Maybe it could be possible to let the adapters handle_info the DOWN messages?

Related Ch issues / PRs: https://github.com/plausible/ch/issues/210, https://github.com/plausible/ch/pull/211

josevalim:
If I understand correctly, the issue is that DBConnection is not very kind to changing the connection once a query starts, right? So we would need to add a mechanism that allows the adapter to say: "try again with another connection, I am going away"?

ruslandoga:
I think we can already do this without "telling" DBConnection like in https://github.com/plausible/ch/pull/211. 

> So we would need to add a mechanism that allows the adapter to say: "try again with another connection, I am going away"?

Do you mean a new connection entirely, or another connection from the pool? The approach I linked above is about a new connection entirely, and that seems to work but kind of feels like the reconnection can happen earlier, not on ping or query attempt. If you mean another connection from the existing pool, they might be disconnected as well.

I was thinking more in terms of:
- connection is idle, it's some time away from the next ping and there is no user query incoming
- underlying TCP connections gets closed
- connection gets DOWN message from the monitor and asks DBConnection to reconnect (via {:disconnect, conn})
- error is logged, making it clear that there might be some instability in the user's system
- the next arriving query gets a connected connection

What seems to be happening right now in Ch is the following:
- connection is idle, it's some time away from the next ping and there is no user query incoming
- underlying TCP connections gets closed
- DBConnection query arrives, uses a closed connection, gets :closed` error back, disconnects, users gets an error that could've been avoided

For some additional context, I think it might be similar to https://github.com/sneako/finch/issues/272

josevalim:
> I was thinking more in terms of

Makes sense! I think this can all be implemented in the connection then, no? Without a need to change DBConnection? You should be able to receive the `:inet.monitor` messages in your connection process.

ruslandoga:
[...]

PR #316, opened 2024-11-11
Proxy info messages to the adapter

ruslandoga:
Context: https://github.com/elixir-ecto/db_connection/issues/315#issuecomment-2467547996

### DIFF
--- a/examples/tcp_connection/lib/tcp_connection.ex
+++ b/examples/tcp_connection/lib/tcp_connection.ex
@@ -148,8 +148,16 @@ defmodule TCPConnection do
   # The handle_info callback is optional and can be removed if not needed.
   # Here it is used to react to `:inet.monitor/1` messages which arrive
   # when socket gets closed while the connection is idle.
-  def handle_info({:DOWN, _ref, _type, sock, _info}, {sock, _buffer} = s) do
-    {:disconnect, TCPConnection.Error.exception({:idle, :closed}), s}
+  def handle_info({:DOWN, _ref, _type, sock, _info}, {sock, _buffer}) do
+    {:disconnect, TCPConnection.Error.exception({:idle, :closed})}
+  end
+
+  def handle_info(msg, state) do
+    Logger.info(fn ->
+      ["#{__MODULE__} (", inspect(self()), ") missed message: ", inspect(msg)]
+    end)
+
+    :ok
   end
 
   @impl true
--- a/examples/tcp_connection/lib/tcp_connection.ex
+++ b/examples/tcp_connection/lib/tcp_connection.ex
@@ -64,6 +64,8 @@ defmodule TCPConnection do
 
     case :gen_tcp.connect(host, port, socket_opts, timeout) do
       {:ok, sock} ->
+        # Monitor the socket so we can react to it being closed. See handle_info/2.
+        _ref = :inet.monitor(sock)
         {:ok, {sock, <<>>}}
 
       {:error, reason} ->
@@ -141,6 +143,13 @@ defmodule TCPConnection do
       {:error, reason} ->
         {:disconnect, TCPConnection.Error.exception({:recv, reason}), state}
     end
+  end
+
+  # The handle_info callback is optional and can be removed if not needed.
+  # Here it is used to react to `:inet.monitor/1` messages which arrive
+  # when socket gets closed while the connection is idle.
+  def handle_info({:DOWN, _ref, _type, sock, _info}, {sock, _buffer} = s) do
+    {:disconnect, TCPConnection.Error.exception({:idle, :closed}), s}
   end
 
   @impl true
--- a/integration_test/cases/info_test.exs
+++ b/integration_test/cases/info_test.exs
@@ -57,4 +57,6 @@ defmodule InfoTest do
              connect: _
            ] = A.record(agent)
   end
+
+  test "handle_info's disconnect with connection checked out"
 end
--- a/integration_test/cases/info_test.exs
+++ b/integration_test/cases/info_test.exs
@@ -5,7 +5,7 @@ defmodule InfoTest do
   alias TestAgent, as: A
   alias TestQuery, as: Q
 
-  test "handle_info handles message and moves on" do
+  test "handle_info handles harmless message and moves on" do
     stack = [
       fn opts ->
         send(opts[:parent], {:connected, self()})
@@ -37,7 +37,7 @@ defmodule InfoTest do
         send(opts[:parent], {:connected, self()})
         {:ok, :state}
       end,
-      {:disconnect, :reason},
+      {:disconnect, RuntimeError.exception("TCP connection just closed")},
       :ok,
       fn opts ->
         send(opts[:parent], :reconnected)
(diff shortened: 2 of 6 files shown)


lib/db_connection/connection.ex
```diff
@@ -330,12 +330,22 @@ defmodule DBConnection.Connection do
...
 
-  def handle_event(:info, msg, :no_state, %{mod: mod} = s) do
-    Logger.info(fn ->
-      [inspect(mod), ?\s, ?(, inspect(self()), ") missed message: " | inspect(msg)]
-    end)
+  def handle_event(:info, msg, :no_state, %{mod: mod, state: state} = s) do
+    if function_exported?(mod, :handle_info, 2) do
+      case apply(mod, :handle_info, [msg, state]) do
+        {:ok, state} ->
+          pool_update(state, s)
```
ruslandoga, line 337: I copied this from `ping`.

lib/db_connection/connection.ex
```diff
@@ -330,12 +330,22 @@ defmodule DBConnection.Connection do
...
+
+        {:disconnect, err, state} ->
+          {:keep_state, %{s | state: state}, {:next_event, :internal, {:disconnect, {:log, err}}}}
```
ruslandoga, line 340: And this too.

lib/db_connection/connection.ex
josevalim: I don't think we need to do a pull update because we don't check the connection out. When we do a ping, [we explicitly check it out from the pool](https://github.com/elixir-ecto/db_connection/blob/master/lib/db_connection/holder.ex#L220), but I don't think we want that here. It should just be a regular message that we don't fiddle with.

ruslandoga: Does it mean we can't update the state from `handle_info/2` or just that we can return the new state without `pool_update/2`?

josevalim: Good question.

I think we either need to not allow state updates or we need to check it out to avoid race conditions. Because the issue is: what happens if we receive this message when the connection is currently checked out? When they are checked out, the state is in the client.

Another option is to check if the connection is checked out or not. If it is checked out, we queue those messages until they are checked in. Then we execute them and check it in. If it is checked in, then we to check them out, execute it, and check them in.

Given you want this exclusively for disconnect, maybe we can get away with no state update for now, but we need to check what happens if we disconnect while checked out. The client will most likely crash (which I assume is what happens nowadays anyway).

ruslandoga: > Given you want this exclusively for disconnect, maybe we can get away with no state update for now, but we need to check what happens if we disconnect while checked out. The client will most likely crash (which I assume is what happens nowadays anyway).

Right, I'll update it to `handle_info/1` with only two possible return values `:ok | :disconnect`. This would be totally enough for Ch. And yes, since catching disconnects early is best effort anyway, letting client who has already checked out the connection crash is fine.

josevalim: I would probably do `{:disconnect, reason}` but looks good to me!

lib/db_connection/connection.ex
```diff
@@ -333,11 +333,11 @@ defmodule DBConnection.Connection do
   def handle_event(:info, msg, :no_state, %{mod: mod, state: state} = s) do
     if function_exported?(mod, :handle_info, 2) do
       case apply(mod, :handle_info, [msg, state]) do
```
ruslandoga, line 335: I'm still passing the state to the callback since the monitor ref the monitor ref / socket ref would be there, and they are probably needed to match on the incoming message.

examples/tcp_connection/lib/tcp_connection.ex
```diff
@@ -148,8 +148,16 @@ defmodule TCPConnection do
   # The handle_info callback is optional and can be removed if not needed.
   # Here it is used to react to `:inet.monitor/1` messages which arrive
   # when socket gets closed while the connection is idle.
-  def handle_info({:DOWN, _ref, _type, sock, _info}, {sock, _buffer} = s) do
-    {:disconnect, TCPConnection.Error.exception({:idle, :closed}), s}
+  def handle_info({:DOWN, _ref, _type, sock, _info}, {sock, _buffer}) do
+    {:disconnect, TCPConnection.Error.exception({:idle, :closed})}
```
ruslandoga, line 152: Like here.

integration_test/cases/info_test.exs
```diff
@@ -1,5 +1,60 @@
...
+
+    assert [
+             connect: _,
+             handle_info: _,
+             handle_status: _,
+             handle_status: _
+           ] = A.record(agent)
+  end
+
+  test "handle_info can force disconnect" do
```
ruslandoga, line 32: I still need to add a test where the message arrives when the client has checked out the connection.

integration_test/cases/info_test.exs
```diff
@@ -0,0 +1,60 @@
...
+      {:idle, :state},
+      {:idle, :state}
+    ]
+
+    {:ok, agent} = A.start_link(stack)
+    {:ok, pool} = P.start_link(agent: agent, parent: self())
+
+    assert_receive {:connected, conn}
+    send(conn, "some harmless message")
+    assert P.run(pool, fn _ -> :result end) == :result
```
ruslandoga, line 22: I'm using this just to "sync" with the connection process, i.e. wait once it processed all the messages. Maybe there is a better way? I usually use `:sys.get_state/1` but don't know who well it would work with DBConnection.

josevalim:

ruslandoga:
Seems like I forgot to run formatter on `examples/` directory.

josevalim:
I just did it in main, no worries. But we also forgot to document `handle_info` as a callback and list it as optional. Can you please send a PR to add that too? Sorry.

ruslandoga:
Hm... I tried `handle_info/2` out in https://github.com/plausible/ch/pull/212 and something seems off. `DBConnection.execute` gets an old connection even though it happens strictly after reconnect. I'll look into it more tomorrow.

```
connect: #Port<0.10>
handshake (in connect): "select 1" on #Port<0.10>
-- first spawn --
handle_execute: "select 1 + 1" on #Port<0.10>
handle_info: #Port<0.10> closed :: {:disconnect, %Mint.TransportError{reason: :closed}}
disconnect: #Port<0.10>
connect: #Port<0.12>
handshake (in connect): "select 1" on #Port<0.12>
-- second spawn --
handle_execute: "select 1 + 1" on #Port<0.10> <-- old conn
```

ruslandoga:
👋 @josevalim 

It doesn't seem like `handle_info` callback does anything. I think I will need to explore other solutions. Should I revert this PR?

josevalim:
@ruslandoga it is up to you. Couldn't we allow `:disconnect` to be returned from `handle_info` and then have that remove it from the table? Another option is to do a checkout, like we do on ping, and then it will be properly disconnected. But in such case we need to be very clear that `handle_info` in our case is a relatively expensive operation, as it checks out, and it must be used only in cases like this.

ruslandoga:
ReqCH gave me an idea that maybe DBConnection is not necessary for a ClickHouse client. I'd like to give an alternative pool a try :)

And it doesn't seem like any other adapter needs a `handle_info` since that issue hasn't been brought up before. So I think I'll revert it to avoid complicating DBConnection unnecessarily.

PR #320, opened 2025-01-13
Use stop instead of raise on genstatem

TylerWitt:
Raise requires Elixir 1.17.2 [in order to not crash the logger](https://github.com/elixir-ecto/postgrex/pull/723#issuecomment-2586781830).

`:stop` allows us to use Elixir's logger instead of the Erlang handling.


josevalim:
I'd prefer to not do this change because it may change semantics in subtle ways (for example, regarding terminate callback). I'd prefer to keep the exception semantics and experience, given that's what happened in the first place. :)

PR #330, opened 2025-09-09
Use process labels

nathanl:
This probably needs tweaking, but I periodically see people asking (and have wondered myself) how to figure out which test in a suite is responsible for something like

```
18:09:26.437 [error] Postgrex.Protocol (#PID<0.413.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.2759.0> exited
```

This is an attempt to make our systems and their errors easier to understand.

### DIFF
--- a/lib/db_connection/util.ex
+++ b/lib/db_connection/util.ex
@@ -1,0 +1,29 @@
+defmodule DBConnection.Util do
+  @moduledoc """
+  Shared functions.
+  """
+
+  @doc """
+  Inspect a pid, including the process label if possible.
+  """
+  def inspect_pid(pid) do
+    case get_label(pid) do
+      :undefined -> inspect(pid)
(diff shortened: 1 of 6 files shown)


lib/db_connection/util.ex
```diff
@@ -0,0 +1,41 @@
...
+  Inspect a pid, including the process label if possible.
+  """
+  def inspect_pid(pid) do
+    with :undefined <- get_label(pid),
+         :undefined <- get_name(pid) do
+      inspect(pid)
+    else
+      label_or_name -> "#{inspect(pid)} (#{inspect(label_or_name)})"
+    end
+  end
```
josevalim, line 14:
```suggestion
  end
  
  defp get_name(pid) do
    try do
      Process.info(pid, :register_name)
    rescue
      _ -> :undefined
    else
      [] -> :undefined
      name -> name
    end
  end
```

PR #331, opened 2025-10-22
Show pool mode in this error

nathanl:
With the current code, it's always :manual when this error is generated, but this is useful information if the user is attempting to use another mode.


josevalim:

PR #332, opened 2025-10-28
Add label (repo name) to ownership errors

nathanl:
A label like `MyApp.Repo` can be passed via the `:label option`. It will be
used in the process label for the `DBConnection.Ownership.Manager` process
and will appear in some error messages. This is helpful
when applications use multiple repos (e.g., read replicas).

Before:

```
  (DBConnection.OwnershipError) cannot find ownership process for #PID<...>
```

After:

```
  (DBConnection.OwnershipError) cannot find ownership process for #PID<...>
  (MyApp.Repo) using mode :manual.
```

Tests verify labels appear correctly in both error types, and that errors
work correctly if no label is given.

In https://github.com/elixir-ecto/ecto_sql/pull/698 the adapter is
updated to pass the repo name as a label.

### DIFF
--- a/lib/db_connection/util.ex
+++ b/lib/db_connection/util.ex
@@ -32,7 +32,7 @@ defmodule DBConnection.Util do
   """
   def set_label(label) do
     if function_exported?(Process, :set_label, 1) do
-      Process.set_label(label)
+      apply(Process, :set_label, [label])
     else
       :ok
     end


josevalim:
Hi @nathanl. I would prefer to not depend on repo, because it is really not a db_connection configuration or concern. Can you find anything on `Process.info(self())` or `Process.info(self(), :registered_name)` you could use?

realglebivanov:
@nathanl would you consider adding some libary-agnostic metadata(contained e.g. in a kw-list) to the exception a viable option?

realglebivanov:
> @realglebivanov Can you elaborate a bit on what that might look like?

So, right now `DBConnection.ConnectionError` contains only message, reason and severity.
Instead of modifying the existing message text, we could add an additional field to the `DBConnection.ConnectionError` that could contain useful metadata.
Also, in this case we wouldn't have to rely on specific option types, it could be just a keyword list.
Of course, that could cost us an overhead of storing and copying around the list.
Hopefully, I made my point clear.

lib/db_connection/ownership/manager.ex
```diff
@@ -107,7 +107,8 @@ defmodule DBConnection.Ownership.Manager do
        mode: mode,
        mode_ref: nil,
        ets: ets,
-       log: log
+       log: log,
+       label: pool_opts[:label]
```
josevalim: Since we are passing the label, wouldn't make more sense to use it for Process.set_label? And then we don't. need to pass it around? It just becomes part of our Util.inspect_pid ?

nathanl:
Still working on this

PR #337, opened 2025-11-09
Allow client to reconnect when specified

josevalim:


josevalim:
@greg-rychlewski here are the changes. The idea is that, when PostgresSQL detects the connection was dropped, we can now return `:disconnect_and_retry`. However, notice we cannot always do this, as a connection may drop in the middleware of a query, and we could retry something dangerous. Therefore, we can only emit retry when preparing/binding queries or issuing begin statements and so on. I am starting with a checkout retries of 3 but for the Neon case we may want to set it to the pool size + 1.

PR #338, opened 2025-11-12
More label usage

nathanl:
To improve error messages and facilitate debugging.

Whenever possible, have error messages that include a PID also include
the process label, name, or initial call.

### DIFF
--- a/lib/db_connection/util.ex
+++ b/lib/db_connection/util.ex
@@ -55,7 +55,7 @@ defmodule DBConnection.Util do
 
   defp get_initial_call(pid) do
     case Process.info(pid, :initial_call) do
-      {:initial_call, _} = tuple -> tuple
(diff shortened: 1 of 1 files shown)


lib/db_connection/holder.ex
```diff
@@ -332,7 +334,7 @@ defmodule DBConnection.Holder do
 
     call_reason =
       if maybe_pid do
-        "Error happened when attempting to transfer to #{inspect(maybe_pid)} " <>
+        "Error happened when attempting to transfer to #{Util.inspect_pid(maybe_pid)} " <>
```
nathanl: Note that `Util.inspect_pid/1` was updated to fall back to `inspect/1` where it's not actually a pid

josevalim:
Thank you, please fix the tests to be less specific and we should be good to go!

nathanl:
@josevalim see update

PR #342, opened 2026-03-25
Fix owner_unallow using wrong ref when switching owners

coop:
When ownership_allow is called with unallow_existing: true to move a process from one owner to another, owner_unallow was removing the process from the caller's owner ref rather than the process's current owner ref. This left the process in the old owner's allowed list.

When the old owner later exits, owner_down removes all processes in its allowed list from checkouts — including the process that was already re-allowed on a different owner. This causes an OwnershipError on the next query because the process is no longer in checkouts.

The fix looks up the process's current checkout entry to find its actual ref and proxy, then cleans up the correct owner's allowed list.

Also fixes the ETS delete in owner_unallow which was passing {pid, proxy} as the key instead of just pid — a no-op on a :set table keyed by the first element.

### DIFF
--- a/integration_test/ownership/manager_test.exs
+++ b/integration_test/ownership/manager_test.exs
@@ -697,6 +697,62 @@ defmodule ManagerTest do
     end
   end
 
+  test "unallow_existing cleans up old owner so its exit doesn't revoke the new allowance" do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #343, opened 2026-04-20
Fix Watcher blocking during slow pool termination

v0idpwn:
The Watcher calls DynamicSupervisor.terminate_child/2 synchronously,
blocking all other watch/terminate operations until the pool finishes
shutting down.

This can cause applications running multiple pools to have degraded
performance when connections take long to terminate (observed in production
with Postgrex).

Added a test reproducing the issue, and a proposed fix via a Task calling
`GenServer.stop/1`.

### DIFF
--- a/lib/db_connection/watcher.ex
+++ b/lib/db_connection/watcher.ex
@@ -38,13 +38,12 @@ defmodule DBConnection.Watcher do
   def handle_info({:DOWN, ref, _, _, _}, {caller_refs, started_refs}) do
     case caller_refs do
       %{^ref => {_supervisor, started_pid, started_ref}} ->
-        Task.Supervisor.start_child(DBConnection.Task, fn ->
-          try do
-            GenServer.stop(started_pid, :shutdown, :infinity)
-          catch
-            :exit, _ -> :ok
-          end
-        end)
+        try do
+          :sys.terminate(started_pid, :shutdown, :infinity)
+        catch
+          :exit, {:noproc, {:sys, :terminate, _}} -> :ok
+          :exit, {:shutdown, {:sys, :terminate, _}} -> :ok
+        end
 
         caller_refs = Map.delete(caller_refs, ref)
         started_refs = Map.put(started_refs, started_ref, {nil, nil})
--- a/lib/db_connection/watcher.ex
+++ b/lib/db_connection/watcher.ex
@@ -45,6 +45,7 @@ defmodule DBConnection.Watcher do
             :exit, _ -> :ok
           end
         end)
+
         caller_refs = Map.delete(caller_refs, ref)
         started_refs = Map.put(started_refs, started_ref, {nil, nil})
         {:noreply, {caller_refs, started_refs}}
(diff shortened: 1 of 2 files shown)


lib/db_connection/watcher.ex
```diff
@@ -56,7 +69,7 @@ defmodule DBConnection.Watcher do
 
   @impl true
   def terminate(_, {_, started_refs}) do
-    for {_, {caller_pid, _}} <- started_refs do
+    for {_, {caller_pid, _}} when caller_pid != nil <- started_refs do
```
v0idpwn, line 71: This was necessary because now we keep `ref => {nil, nil}` when a termination is on flight

lib/db_connection/watcher.ex
```diff
@@ -37,16 +37,29 @@ defmodule DBConnection.Watcher do
   @impl true
   def handle_info({:DOWN, ref, _, _, _}, {caller_refs, started_refs}) do
     case caller_refs do
-      %{^ref => {supervisor, started_pid, started_ref}} ->
-        Process.demonitor(started_ref, [:flush])
-        DynamicSupervisor.terminate_child(supervisor, started_pid)
-        {:noreply, {Map.delete(caller_refs, ref), Map.delete(started_refs, started_ref)}}
+      %{^ref => {_supervisor, started_pid, started_ref}} ->
+        Task.Supervisor.start_child(DBConnection.Task, fn ->
```
josevalim: Instead of spawning at task, we could call sys:terminate and await for the down message. This is how GenServer.stop is implemented:

```erlang
stop(Process, Reason, Timeout) ->
    Mref = erlang:monitor(process, Process),
    T0 = erlang:monotonic_time(millisecond),
    StopTimeout = fun(infinity) -> infinity;
                     (T1) -> T1 - (((erlang:monotonic_time(microsecond) + 999) div 1000) - T0)
                  end,
    RemainingTimeout =
        try sys:terminate(Process, Reason, Timeout) of
            ok ->
                StopTimeout(Timeout)
        catch
            exit:{noproc, {sys, terminate, _}} ->
                demonitor(Mref, [flush]),
                exit(noproc);
            exit:{timeout, {sys, terminate, _}} ->
                demonitor(Mref, [flush]),
                exit(timeout);
            exit:{Reason, {sys, terminate, _}} ->
                StopTimeout(Timeout);
            exit:Reason1 ->
                demonitor(Mref, [flush]),
                exit(Reason1)
        end,
    receive
        {'DOWN', Mref, _, _, Reason} ->
            ok;
        {'DOWN', Mref, _, _, Reason2} ->
            exit(Reason2)
    after RemainingTimeout ->
        demonitor(Mref, [flush]),
        exit(timeout)
    end.
```

Since we don't care about the timeout (it is `:infinity`), a lot of the code above could be simplified. It is mostly sys:terminate in a try/catch + receive!

v0idpwn: Sounds good! TIL `gen_server:stop` uses `sys:terminate` underneath 🤯

josevalim (approved):

v0idpwn:
Thank you @josevalim! 💚 
We tested it with Postgrex and it seems to have resolved the issue. We will be rolling it out in production and I'll report back confirming the results.

PR #345, opened 2026-04-25
Fix max_lifetime docstring inaccuracies, validation, and typos

ericmj:
- docstring: replace `idle_limit` (a count) with `2 * idle_interval` in the worst-case idle-wait formula; the addition was dimensionally incorrect.
- docstring: drop "and on checkout" — `conn(:ts)` is set only at connection establishment, never reset on checkout.
- docstring: fix "you ca set it" typo.
- connection_pool: reject negative or descending `:max_lifetime` ranges (e.g. `100..50//1`) at init instead of letting them flow into `:erlang.phash2/2` and be silently swallowed by the rescue in `Holder.maybe_disconnect/4`.
- holder: rename misspelled `ellapsed` to `elapsed`.

### DIFF
--- a/lib/db_connection.ex
+++ b/lib/db_connection.ex
@@ -440,12 +440,12 @@ defmodule DBConnection do
 
     * `:max_lifetime` - The number of ms the connection is allowed to live.
       It is a range so you can jitter/spread disconnections over some time period.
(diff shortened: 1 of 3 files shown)


josevalim:

PR #347, opened 2026-05-08
Use connection establishment time, not the last ping, for `max_lifetime` calculations

PeterCxy:
The holder's ts is reset on every update, which happens on pings too. This renders `max_lifetime` often useless since along with the default `idle_interval` the timestamp is reset every second.

This PR makes `max_lifetime` logic use connection establishment time, which is passed along on holder updates. The new test fails without changes made in this PR.

### DIFF
--- a/integration_test/connection_pool/max_lifetime_test.exs
+++ b/integration_test/connection_pool/max_lifetime_test.exs
@@ -1,3 +1,23 @@
+defmodule MaxLifetimeTest.Connection do
+  def connect(opts) do
+    {:ok, %{parent: Keyword.fetch!(opts, :parent), pings: 0}}
+  end
+
+  def disconnect(err, state) do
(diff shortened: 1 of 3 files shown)


josevalim:
@PeterCxy good find! Perhaps instead of storing the connection_at in the holder, we can pass it on the timestamp?

PeterCxy:
Yeah that would for sure make sense too, I was just not sure whether `disconnect_all_reason` expects the ts to _actually_ be the timestamp of the last ping since I didn't look into what the other function is for :smile: I'll look more into this next week.

josevalim:
@PeterCxy it is meant to be the connected_at time, because it is used for a similar feature. So the previous patch was "right" but there was this hidden bug in `ts`. I renamed it to `connected_at` (as in this PR) and pushed it to main, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #348, opened 2026-05-22
fix: include pool label in timeout errors

dl-alexandre:
## Summary

DBConnection already receives a pool `:label` from callers such as Ecto SQL (`connection.child_spec([label: name] ++ config)`). This PR uses that existing label in timeout diagnostics so applications with multiple Repo pools can identify which pool exhausted its queue or held a checkout too long.

Without the label, timeout errors only say that a client queued, checked out, or owned "the connection" for too long. In an Ecto application with several repos or partitioned pools, that is not enough information to know which Repo needs attention.

## Changes

- labels default connection pool and ownership proxy processes when `:label` is provided
- includes the label in checkout, queue-drop, and ownership timeout messages as `connection from MyApp.Repo`
- documents `:label` as a DBConnection start option and includes it in `available_start_options/0`
- adds behavior-level coverage for the queue-drop timeout message instead of only testing the formatting helper

## Ecto behavior

Ecto SQL already passes the repo name as the DBConnection label when building the connection child spec. With this change, a timeout from an Ecto repo pool can point back to the configured repo name in the DBConnection error message.

## Verification

- `mix format --check-formatted`
- `mix test test/db_connection_test.exs integration_test/cases/queue_test.exs`

### DIFF
--- a/integration_test/cases/queue_test.exs
+++ b/integration_test/cases/queue_test.exs
@@ -42,6 +42,29 @@ defmodule QueueTest do
     end
   end
 
+  test "queue drop timeout includes pool label when present" do
+    if function_exported?(:proc_lib, :get_label, 1) do
+      stack = [{:ok, :state}, {:idle, :state}, {:idle, :state}]
+      {:ok, agent} = A.start_link(stack)
+
+      opts = [
+        agent: agent,
+        parent: self(),
+        label: TestRepo,
+        queue_target: 10,
+        queue_interval: 10
+      ]
+
+      {:ok, pool} = P.start_link(opts)
+
+      P.run(pool, fn _ ->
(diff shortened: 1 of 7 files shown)


josevalim:
There is no description on what is the benefit and why this should be merged, so it is unclear why to go ahead with this pull request.

josevalim:
How is the label being set? How will it work with Ecto? This is perhaps the second or third time we ask you to be careful with pull requests if you are vibe coding them. Please be mindful of our time.

dl-alexandre:
Thanks, that feedback is fair. I updated the PR description to explain the benefit and the Ecto path, and pushed a follow-up that documents `:label`, improves the message wording, and adds behavior-level coverage for the queue-drop timeout path.

PR #357, opened 2026-07-19
Reject checkouts after deadline

### LINKED ISSUE #356: Race conditions leading to silent drop of expired deadlines
`DBConnection.Holder.checkout_call/5` starts the deadline timer before storing its reference in the holder’s ETS record. The pool may receive the timeout immediately when the absolute deadline has already elapsed and discard it (`:deadline` field is still `nil`). The checked out connection has no deadline and may starve the pool.

The chances of this happening in real life are slim but this script reproduces it via explicitly passing expired deadline

Repro:
```elixir
# Run with: elixir repro_expired_deadline.exs
#
# By default this installs the affected 2.10.2 release. To check a local fix:
# DB_CONNECTION_PATH=. elixir repro_expired_deadline.exs

dependency =
  case System.get_env("DB_CONNECTION_PATH") do
    nil -> {:db_connection, "== 2.10.2"}
    path -> {:db_connection, path: path}
  end

Mix.install([dependency])
Logger.configure(level: :critical)

defmodule Repro.Connection do
  @behaviour DBConnection

  @impl true
  def connect(_opts), do: {:ok, nil}

  @impl true
  def disconnect(_error, _state), do: :ok

  @impl true
  def checkout(state), do: {:ok, state}

  @impl true
  def ping(state), do: {:ok, state}

  @impl true
  def handle_status(_opts, state), do: {:idle, state}

  @impl true
  def handle_begin(_opts, state), do: {:ok, :begun, state}

  @impl true
  def handle_commit(_opts, state), do: {:ok, :committed, state}

  @impl true
  def handle_rollback(_opts, state), do: {:ok, :rolled_back, state}

  @impl true
  def handle_prepare(query, _opts, state), do: {:ok, query, state}

  @impl true
  def handle_execute(query, _params, _opts, state), do: {:ok, query, :result, state}

  @impl true
  def handle_close(_query, _opts, state), do: {:ok, :closed, state}

  @impl true
  def handle_declare(query, _params, _opts, state), do: {:ok, query, :cursor, state}

  @impl true
  def handle_fetch(_query, _cursor, _opts, state), do: {:halt, :done, state}

  @impl true
  def handle_deallocate(_query, _cursor, _opts, state), do: {:ok, :done, state}
end

defmodule Repro do
  @attempts 100

  def run do
    {:ok, pool} =
      DBConnection.start_link(Repro.Connection,
        pool: DBConnection.ConnectionPool,
        pool_size: 1,
        backoff_min: 0
      )

    case reproduce(pool, 1) do
      {:reproduced, attempt, deadline} ->
        IO.puts("BUG REPRODUCED on attempt #{attempt}")
        IO.puts("expired deadline: #{deadline}")
        IO.puts("callback entered: true")
        IO.puts("expected: DBConnection.ConnectionError before callback entry")

      :not_reproduced ->
        IO.puts("Bug not reproduced in #{@attempts} attempts")
        System.halt(1)
    end
  end

  defp reproduce(_pool, attempt) when attempt > @attempts, do: :not_reproduced

  defp reproduce(pool, attempt) do
    deadline = System.monotonic_time(:millisecond) - 1
    marker = make_ref()

    try do
      DBConnection.run(
        pool,
        fn _connection ->
          send(self(), marker)
          :ok
        end,
        deadline: deadline
      )
    rescue
      DBConnection.ConnectionError -> :deadline_rejected
    catch
      :exit, _reason -> :connection_exited
    end

    receive do
      ^marker -> {:reproduced, attempt, deadline}
    after
      0 ->
        Process.sleep(1)
        reproduce(pool, attempt + 1)
    end
  end
end

Repro.run()
```

Result:
```
BUG REPRODUCED on attempt 1
expired deadline: -576460751577
callback entered: true
expected: DBConnection.ConnectionError before callback entry
```

lukaszsamson:
Fixes #356

### DIFF
--- a/integration_test/cases/queue_test.exs
+++ b/integration_test/cases/queue_test.exs
@@ -86,6 +86,29 @@ defmodule QueueTest do
     assert P.run(pool, fn _ -> :result end) == :result
   end
 
+  test "does not checkout a connection after its deadline" do
+    stack = [
+      {:ok, :state},
+      :ok,
+      fn opts ->
+        send(opts[:parent], :reconnected)
+        {:ok, :state}
+      end
+    ]
+
+    {:ok, agent} = A.start_link(stack)
+
+    opts = [agent: agent, parent: self()]
+    {:ok, pool} = P.start_link(opts)
+    deadline = System.monotonic_time(:millisecond) - 1
+
+    assert_raise DBConnection.ConnectionError, ~r/deadline reached/, fn ->
+      P.run(pool, fn _ -> flunk("checkout succeeded") end, deadline: deadline)
+    end
+
+    assert_receive :reconnected
+  end
+
   test "queue many async exits" do
     stack = [{:ok, :state}] ++ List.duplicate({:idle, :state}, 20)
     {:ok, agent} = A.start_link(stack)
--- a/integration_test/ownership/owner_test.exs
+++ b/integration_test/ownership/owner_test.exs
@@ -132,7 +132,7 @@ defmodule OwnerTest do
                fn _ ->
                  assert_receive :reconnected
                end,
-               timeout: 0
+               timeout: 10
              )
            end) =~ ~r"timed out"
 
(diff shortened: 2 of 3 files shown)


lukaszsamson:
@josevalim Would you like any changes to this PR?

josevalim:
thanks, it fell through the cracks!

